PR #1, opened 2021-03-24
Main cli with voters rewards command

danielSanchezQ:


danielSanchezQ:
We require some poldercast dependencies to be updated. Currently is not compiling.

danielSanchezQ:
I'm working on making it work with fixed-point numbers instead.

catalyst-toolbox-cli/src/cli/rewards/voters.rs
```diff
@@ -0,0 +1,123 @@
...
+    stake_per_voter
+        .iter()
+        .map(|(k, v)| (*k, total_stake.div(v)))
+        .collect()
+}
+
+/// caculate total reward from the inverse of the the reward share
+fn calculate_total_reward(share: u64, total_reward: u64) -> f64 {
+    (total_reward as f64).div(share as f64)
+}
```
danielSanchezQ, line 64: @eugene-babichenko I did this tiny trick where I store the inverse of the reward share and then I just use the f64 in the last step. This should keep the numbers as small as possible in the last calculation and shouldn't be any problem. Data looks good for latest fund.

catalyst-toolbox-cli/src/cli/rewards/voters.rs
```diff
@@ -0,0 +1,123 @@
...
+    csv_writer.write_record(&header).map_err(Error::Csv)?;
+
+    for (address, reward) in results.iter() {
+        let stake = stake_per_voter.get(*address).unwrap();
+        let voter_reward = calculate_total_reward(*reward, total_rewards);
+        let record = [
+            address.to_string(),
+            stake.to_string(),
+            format!("{:.0}", voter_reward.trunc()),
+            format!("{:.4}", voter_reward.mul(ADA_TO_LOVELACE_FACTOR as f64)),
```
enzoc4, line 89: I don't think the amount in lovelaces can have floating point digits, so it's a bit weird to format it that way. (I mean, lovelaces are the minimal unit of representation, the conversion is lossless)

catalyst-toolbox-cli/src/cli/rewards/voters.rs
```diff
@@ -0,0 +1,123 @@
...
+        "Stake of the voter (ADA)",
+        "Reward for the voter (ADA)",
+        "Reward for the voter (lovelace)",
+    ];
+    let mut csv_writer = csv::Writer::from_writer(writer);
+    csv_writer.write_record(&header).map_err(Error::Csv)?;
+
+    for (address, reward) in results.iter() {
+        let stake = stake_per_voter.get(*address).unwrap();
+        let voter_reward = calculate_total_reward(*reward, total_rewards);
```
enzoc4, line 84: just nitpicking, but I didn't expect to find a `calculate` function inside the `write` one, I was expecting to see it before.

Although in this case it's just one line, so I guess it makes sense

catalyst-toolbox-cli/src/cli/rewards/voters.rs
```diff
@@ -0,0 +1,123 @@
...
+fn write_rewards_results(
+    common: Common,
+    stake_per_voter: &HashMap<&Address, u64>,
+    results: &HashMap<&Address, u64>,
+    total_rewards: u64,
+) -> Result<(), Error> {
+    let writer = common.open_output()?;
+    let header = [
+        "Address",
+        "Stake of the voter (ADA)",
```
enzoc4, line 75: Also, I think this is in lovelaces. Unless you are really sure this is in ADA, but I doubt the snapshot is in adas.

catalyst-toolbox-cli/src/cli/rewards/voters.rs
```diff
@@ -0,0 +1,123 @@
...
+            Initial::Cert(_) => {}
+            Initial::LegacyFund(_) => {}
+        }
+    }
+    (total_stake, stake_per_voter)
+}
+
+/// Rewards are u64 for keeping the it we would calculate the inverse total_stake/voter_stake
+fn calculate_inverse_reward_share<'address>(
+    total_stake: u64,
```
enzoc4, line 52: maybe use `Value` instead of `u64`? Or maybe some different newtype (`Lovelaces`?), although this is nitpicking too.

enzoc4:
also, I don't see why couldn't we pull a library to make fixed-precision / fractional math? Does this really need to be fast?

catalyst-toolbox-cli/src/cli/rewards/voters.rs
danielSanchezQ: Yeah, I changed this a few times and I had some leftovers.

catalyst-toolbox-cli/src/cli/rewards/voters.rs
danielSanchezQ: Absolutely, it wasn't there before. I tried to keep away from the f64 variant until the last stage so I called it there. Probably I could take it outside.

catalyst-toolbox-cli/src/cli/rewards/voters.rs
danielSanchezQ: I had a mess with this. It is in ADA or Lovelaces? I thought it was in lovelaces first then in ADA and I have been switching between them.

catalyst-toolbox-cli/src/cli/rewards/voters.rs
danielSanchezQ: Yeah, I could use the value wrapper for this. Would make sense.

danielSanchezQ:
@enzoc4 Yeah, it was suggested before too. Any library suggestion?

catalyst-toolbox-cli/src/cli/rewards/voters.rs
enzoc4: Well, I didn't actually check the snapshot code this time, but I always thought it is in lovelaces... It'd be weird if we changed it, why would we throw away the fractional part?

enzoc4:
@danielSanchezQ not really, actually... Never tried one, maybe this one? https://crates.io/crates/rust_decimal
Probably someone else knows more about this.

I said it mostly to be safed, though, I didn't analyze the kind of numbers that we are going to be using, btw.

catalyst-toolbox-cli/src/main.rs
```diff
@@ -0,0 +1,20 @@
...
+
+fn report_error(error: Box<dyn Error>) {
+    eprintln!("{}", error);
+    let mut source = error.source();
+    while let Some(sub_error) = source {
+        eprintln!("  |-> {}", sub_error);
+        source = sub_error.source();
+    }
+    std::process::exit(1)
+}
```
eugene-babichenko, line 20: same here

eugene-babichenko (changes requested):

catalyst-toolbox-cli/src/cli/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl CatalystCommand {
+    pub fn exec(self) -> Result<(), Box<dyn Error>> {
+        use self::CatalystCommand::*;
+        match self {
+            Rewards(rewards) => rewards.exec()?,
+        };
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
danielSanchezQ: Damn, I trust too much in the `rustfmt` 😞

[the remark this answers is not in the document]
eugene-babichenko: Strange that it doesn't fix this problem though. I thought that newline in the end of the file is conventional across almost all programming languages.

eugene-babichenko (approved):
The code itself is looking good. What is left to do is to evaluate the correctness of the computations :)

danielSanchezQ:
Merged this. If anything needs changes I will open a new branch with a fix.

PR #2, opened 2021-03-24
Proposers rewards

danielSanchezQ:
Need of #1 to be merged first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,8 @@ Load your __venv__ and/or run with your default __python3__ `python proposers_re
 ```shell
 Usage: proposers_rewards.py [OPTIONS]
 
-  Calculate catalyst rewards after tallying process. If all --proposals-
-  path, --active-voteplan-path and --challenges_path are provided data is
+  Calculate catalyst rewards after tallying process. If all, --proposals-
+  path --active-voteplan-path and --challenges_path, are provided. Data is
   loaded from the json files on those locations. Otherwise data is requested
   to the proper API endpoints pointed to the --vit-station-url option.
 
--- a/README.md
+++ b/README.md
@@ -72,15 +72,19 @@ Load your __venv__ and/or run with your default __python3__ `python proposers_re
 ```shell
 Usage: proposers_rewards.py [OPTIONS]
 
-  Calculate catalyst rewards after tallying process. If all, --proposals-
-  path --active-voteplan-path and --challenges_path, are provided. Data is
-  loaded from the json files on those locations. Otherwise data is requested
-  to the proper API endpoints pointed to the --vit-station-url option.
+  Calculate catalyst rewards after tallying process. If all --proposals-path, 
+  --active-voteplan-path and --challenges_path are provided. Then,
+  data is loaded from the json files on those locations. Otherwise data is
+  requested to the proper API endpoints pointed to the --vit-station-url
+  option. Rewards are written into a separated file for each challenge. File
+  is constructed via the --output-file. For example /out/rewards.csv with
+  challenges [challenge_1, challenge_2] will generate
+  /out/rewards_challenge_1.csv and /out/rewards_challenge_2.csv files.
 
 Options:
   --conversion-factor FLOAT       [required]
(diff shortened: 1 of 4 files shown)


danielSanchezQ:
Since it was implemented we can already use it. But most probably I'll migrate it to rust and reuse things from `vit-servicing-station-lib` that would be much handier.

README.md
```diff
@@ -59,3 +59,41 @@ OPTIONS:
...
+
+  Calculate catalyst rewards after tallying process. If all --proposals-
+  path, --active-voteplan-path and --challenges_path are provided data is
+  loaded from the json files on those locations. Otherwise data is requested
+  to the proper API endpoints pointed to the --vit-station-url option.
+
+Options:
+  --conversion-factor FLOAT       [required]
+  --output-file TEXT              [required]
+  --threshold FLOAT               [default: 0.15]
```
Zeegomo, line 83: Nit: maybe a more descriptive name like approval threshold?

scripts/python/proposers_rewards.py
```diff
@@ -0,0 +1,416 @@
...
+# File loaders
+
+def load_json_from_file(file_path: str) -> Dict:
+    with open(file_path, encoding="utf-8") as f:
+        return json.load(f)
+
+
+def get_proposals_from_file(proposals_file_path: str) -> Dict[str, Proposal]:
+    proposals: Generator[Proposal, None, None] = (
+        Proposal(**proposal_data) for proposal_data in load_json_from_file(proposals_file_path)
```
Zeegomo, line 121: what is the advantage of using a generator instead of list comprehension?

Zeegomo (approved):
LGTM

scripts/python/proposers_rewards.py
danielSanchezQ: It is not consumed immediately which avoids the intermediate allocation.

README.md
danielSanchezQ: Definitely, thanks :)

scripts/python/proposers_rewards.py
Zeegomo: ahh, so also the dict below is created piece by piece upon request?

danielSanchezQ: Yeah, generators are lazy so it is consumed by the dict below.

scripts/python/proposers_rewards.py
```diff
@@ -0,0 +1,416 @@
...
+class DecryptedTally(pydantic.BaseModel):
+    decrypted: TallyResult
+
+
+class PrivateTallyState(pydantic.BaseModel):
+    state: DecryptedTally
+
+
+class PrivateTally(pydantic.BaseModel):
+    Private: PrivateTallyState
```
enzoc4, line 66: just curious, why is this uppercase? (`Private` instead of `private`)

enzoc4 (approved):

scripts/python/proposers_rewards.py
```diff
@@ -0,0 +1,416 @@
...
+        proposals_result = await client.get(f"{vit_servicing_station_url}/api/v0/proposals")
+        assert proposals_result.status_code == 200
+        return [Proposal(**proposal_data) for proposal_data in proposals_result.json()]
+
+
+async def get_active_voteplans_from_api(vit_servicing_station_url: str) -> List[VoteplanStatus]:
+    async with httpx.AsyncClient() as client:
+        proposals_result = await client.get(f"{vit_servicing_station_url}/api/v0/vote/active/plans")
+        assert proposals_result.status_code == 200
+        return [VoteplanStatus(**proposal_data) for proposal_data in proposals_result.json()]
```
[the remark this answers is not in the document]
danielSanchezQ: Yeah, they can be refactored. I thought about it too, Ill do.

scripts/python/proposers_rewards.py
danielSanchezQ: in the talley result it is defined as uppercase:

```json
"tally": {
          "Public": {
```

PR #4, opened 2021-04-16
Fixed precision error

danielSanchezQ:
Fixed a precision error in the calculations of the voter's rewards.
Updated rewards input to be in Lovelace instead.


danielSanchezQ:
@disasm Would you take a look at it and use it to compare with previous results to double-check it?

mzabaluev (approved):

PR #7, opened 2021-05-03
Rust gh actions

danielSanchezQ:


.github/workflows/release.yml
```diff
@@ -0,0 +1,245 @@
...
+      upload_url: ${{ steps.create_release.outputs.upload_url }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          submodules: 'true'
+
+      - id: release_info
+        name: Get release information
+        run: python3 ci/release-info.py "$GITHUB_EVENT_NAME"
```
Zeegomo, line 30: I think this file is missing, and there is some hard-coded stuff there, like the workspaces in jormungandr

.github/workflows/release.yml
```diff
@@ -0,0 +1,245 @@
...
+        with:
+          toolchain: ${{ matrix.toolchain }}
+          target: ${{ matrix.config.target }}
+          override: true
+          default: true
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          submodules: true
```
Zeegomo, line 149: This is probably not needed?

Zeegomo (approved):

PR #8, opened 2021-05-05
Fragment log recovery

danielSanchezQ:


catalyst-toolbox-cli/src/cli/recovery/tally/fragments.rs
```diff
@@ -0,0 +1,94 @@
...
+        FileFragmentsIterator {
+            reader: self.reader,
+        }
+    }
+}
+
+impl Iterator for FileFragmentsIterator {
+    type Item = PersistentFragmentLog;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
eugene-babichenko, line 39: This is some nice idiomatic Rust here!

One note: I think it is actually fine to have `Result` as `Item`. It is easy to process them in a `for` loop and if you use combinators + `collect` then it is just `.collect::<Result<Vec<_>, _>>`. This gives us proper error reporting at the expense of a few more bytes.

catalyst-toolbox-cli/src/cli/recovery/tally/fragments.rs
```diff
@@ -0,0 +1,94 @@
...
+        let codec = bincode::DefaultOptions::new()
+            .with_fixint_encoding()
+            .allow_trailing_bytes();
+        codec.deserialize_from(&mut self.reader).ok()
+    }
+}
+
+pub fn get_fragments_log_files_path(folder: &Path) -> io::Result<impl Iterator<Item = PathBuf>> {
+    let mut entries: Vec<_> = fs::read_dir(folder)?
```
eugene-babichenko, line 48: A similar note here: I would rather write a `for` loop and return errors from `entry`. I am not even sure when they could appear, probably a permissions problem which is very unlikely in this setup.

catalyst-toolbox-cli/src/cli/recovery/tally/mockchain.rs
```diff
@@ -0,0 +1,86 @@
+use chain_impl_mockchain::block::HeaderId;
+use chain_impl_mockchain::fee::LinearFee;
```
eugene-babichenko, line 2: That's a mild warning, aren't we using `clippy` with `-- -D warnings`?

catalyst-toolbox-cli/src/cli/recovery/tally/mockchain.rs
```diff
@@ -0,0 +1,86 @@
...
+    while !fragments.is_empty() && counter < fragments.len() {
+        counter += 1;
+        match fragments.pop_front() {
+            None => {
+                break;
+            }
+            Some(fragment_log) => {
+                if matches!(&fragment_log.fragment, Fragment::VoteTally(_)) {
+                    tally = Some(fragment_log.fragment.clone());
+                    continue;
```
eugene-babichenko, line 32: Would break when we have multiple vote plans.

catalyst-toolbox-cli/src/cli/recovery/tally/mockchain.rs
```diff
@@ -0,0 +1,86 @@
...
+    block0_header_id: HeaderId,
+    block0_fragments: impl Iterator<Item = &'block0 Fragment>,
+    logged_fragments: impl Iterator<Item = PersistentFragmentLog>,
+) -> Result<(Ledger, VecDeque<PersistentFragmentLog>), chain_impl_mockchain::ledger::Error> {
+    let mut ledger = Ledger::new(block0_header_id, block0_fragments)?;
+    let parameters = ledger.get_ledger_parameters();
+    let block_date = ledger.date();
+    let mut fragments = VecDeque::from_iter(logged_fragments);
+    println!("{}", fragments.len());
+    let mut counter = 0;
```
eugene-babichenko, line 21: What is the idea behind using this counter? I am not sure I follow the logic here.

eugene-babichenko (commented):
We need to do some time jumps here to properly emulate vote and tally periods. Otherwise this is a good start.

catalyst-toolbox-cli/src/cli/recovery/tally/mockchain.rs
Zeegomo: Avoid looping indefinitely if there are some fragments that are not valid (e.g. a spending counter is skipped)

catalyst-toolbox-cli/src/cli/recovery/tally/mockchain.rs
danielSanchezQ: Aha, its true, we would have a tally for each voteplan. Nice catch, thanks!

catalyst-toolbox-cli/src/cli/recovery/tally/mockchain.rs
```diff
@@ -0,0 +1,108 @@
...
+                        fragments.push_back(fragment_log);
+                    }
+                }
+            }
+        };
+    }
+
+    // postprocess failed fragments
+    let mut counter = 0;
+    while !fragments.is_empty() && counter < fragments.len() {
```
enzoc4, line 46: I've been wondering if we should use `retain` here instead, something like... (untested, though)

```rust
    while !fragments.is_empty() {
        let len_before = fragments.len();

        fragments.retain(|fragment_log| {
            ledger
                .apply_fragment(&parameters, &fragment_log.fragment, block_date)
                .is_err()
        });

        if len_before == fragments.len() {
            break;
        }
    }
```

eugene-babichenko: Not a bad idea actually, just need to take care of logging errors here.

catalyst-toolbox-cli/src/cli/recovery/tally/mockchain.rs
```diff
@@ -0,0 +1,108 @@
...
+        match fragment {
+            Err(e) => {
+                println!("Error processing fragment: {:?}", e);
+            }
+            Ok(fragment_log) => {
+                if matches!(&fragment_log.fragment, Fragment::VoteTally(_)) {
+                    tally_fragments.push(fragment_log);
+                    continue;
+                }
+                match ledger.apply_fragment(&parameters, &fragment_log.fragment, block_date) {
```
Zeegomo, line 32: `apply_fragment` returns the new ledger, we need to re-assign it in case of success

Zeegomo: oops, sorry missed those

danielSanchezQ: I haven't pushed them still though 😅

catalyst-toolbox-cli/src/cli/recovery/tally/fragments.rs
mzabaluev: Yes, a fallible iterator is not very elegant, but `collect` and other combinators work on it.

danielSanchezQ:
I came up with a new approach and with @Zeegomo we have implemented a first skeleton of what will it be.
The idea is to process all incoming fragments with new spending counters as they come. We use a mirror architecture is where we map old accounts to new accounts and then recreate transactions for those new accounts using an increasing spending counter. It is still a WIP but it would be much better than the other approach which started to become a big and unstable Frankenstain monster.

mzabaluev:
What's up with the test failure?

Zeegomo:
Oops, we left a path to a local file that we used for testing (and some clippy lints)

danielSanchezQ:
We do not have a current test right now. It was mostly used for manual debugging. I would like to keep it and update it when we have some data to really make a test with. Or we could just remove it entirely.

catalyst-toolbox-cli/src/cli/recovery/tally.rs
```diff
@@ -0,0 +1,86 @@
...
+
+impl Replay {
+    pub fn exec(self) -> Result<(), Error> {
+        let Replay {
+            block0_path,
+            logs_path,
+            output,
+            output_format,
+        } = self;
+        let block0 = read_block0(block0_path)?;
```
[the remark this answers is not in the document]
danielSanchezQ: All needed data should be in block0 already yes.

catalyst-toolbox-cli/src/cli/recovery/tally.rs
```diff
@@ -0,0 +1,86 @@
...
+        let fragments = load_persistent_fragments_logs_from_folder_path(&logs_path)?;
+
+        let (ledger, failed) = recover_ledger_from_logs(&block0, fragments)?;
+        if !failed.is_empty() {
+            println!("{} fragments couldn't be properly processed", failed.len());
```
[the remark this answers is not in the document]
danielSanchezQ: Yeah, forgot about that output feature. Thanks.

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -188,20 +187,20 @@ pub fn recover_ledger_from_logs(
                 let block_date = fragment_log_timestamp_to_blockdate(time, &timeframe, &ledger)
                     .expect("BlockDates should always be valid for logs timestamps");
 
-                println!("Fragment processed {}", fragment.hash());
+                dbg!("Fragment processed {}", fragment.hash());
```
mzabaluev, line 190: This would be good to put behind a `--verbose` command line flag. Not necessary for this PR, though.

danielSanchezQ: I like that idea. Probably to be added later as you suggest.

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -0,0 +1,465 @@
...
+    spending_counter: SpendingCounter,
+    block0_hash: &HeaderId,
+    sign_data_hash: &TransactionSignDataHash,
+    account: &account::Identifier,
+    witness: &account::Witness,
+    range_check: Range<i32>,
+) -> (bool, u32) {
+    let spending_counter: i32 = <u32>::from(spending_counter) as i32;
+    for i in range_check {
+        let new_spending_counter = spending_counter.add(i).clamp(0, i32::MAX) as u32;
```
mzabaluev, line 116: It's probably slower if you start with `spending_counter - 50` if the actual drift tends to be around the expected value. I'd do iteration over `n`, `n + 1`, `n - 1`, ...

Also, using `clamp` means you might do redundant checks. Don't cast to a `i32`, but use something like:

```rust
    if let Some(n) = spending_counter.checked_add(i) {
        // try with n as spending counter
    } else {
        continue;
    }
    if let Some(n) = spending_counter.checked_sub(i) {
        // try with n as spending counter
    } else {
        continue;
    }
```

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -0,0 +1,465 @@
...
+struct FragmentReplayer {
+    wallets: HashMap<Address, Wallet>,
+    spending_counters: HashMap<Address, Vec<u32>>,
+    voteplans: HashMap<VotePlanId, VotePlan>,
+    old_block0_hash: Hash,
+    fees: LinearFee,
+}
+
+impl FragmentReplayer {
+    const CHECK_RANGE: Range<i32> = -50..50;
```
mzabaluev, line 250: The choice of limits needs to be carefully considered. I'd rather use the number of proposals in all vote plans multiplied by two to be absolutely sure. Coupled with the expectation-centered counter picking as detailed in the comment above, this should not be very slow when we expect the majority of ballots to be legitimately created.

danielSanchezQ: Ok, makes sense.

danielSanchezQ: I implemented as the number of proposals range up and range down, it should be enough:
`-proposals <- counter -> +proposals`

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -0,0 +1,502 @@
...
+    use std::io::BufReader;
+    use std::path::PathBuf;
+
+    fn read_block0(path: PathBuf) -> std::io::Result<Block> {
+        let reader = std::fs::File::open(path)?;
+        Ok(Block::deserialize(BufReader::new(reader)).unwrap())
+    }
+
+    #[test]
+    fn test_vote_flow() -> std::io::Result<()> {
```
eugene-babichenko, line 455: Is there a description for the testing data? I think we are interested in whether it has unordered fragments.

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -0,0 +1,502 @@
...
+                let voteplan = tx.as_slice().payload().into_payload();
+                Some((voteplan.to_id(), voteplan))
+            } else {
+                None
+            }
+        })
+        .collect()
+}
+
+fn verify_original_tx(
```
eugene-babichenko, line 114: What is the motivation to do that? An heuristic to check of the fragment should be rejected unconditionally?

eugene-babichenko (commented):
I have a few questions regarding the logic + some formatting nitpicks. The overall idea is looking good. And I trust the passing test ;)

catalyst-toolbox-lib/src/recovery/tally.rs
danielSanchezQ: It is just to check the signature of the content, but without checking the actual spending counter (and instead check a possible range of them).

catalyst-toolbox-lib/src/recovery/tally.rs
danielSanchezQ: There is another PR incoming with better testing which actually tests unordered fragments.

danielSanchezQ:
@dkijania , @eugene-babichenko , I just merged @Zeegomo tests.

catalyst-toolbox-cli/src/cli/recovery/tally.rs
```diff
@@ -0,0 +1,86 @@
...
+        if !failed.is_empty() {
+            eprintln!("{} fragments couldn't be properly processed", failed.len());
+            for failed_fragment in failed {
+                eprintln!("{}", failed_fragment.id());
+            }
+        }
+        let voteplans = ledger.active_vote_plans();
+        let voteplan_status: Vec<VotePlanStatus> =
+            voteplans.into_iter().map(VotePlanStatus::from).collect();
+        let mut out_writter = output.open()?;
```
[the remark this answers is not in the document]
danielSanchezQ: Thanks!

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -0,0 +1,648 @@
...
+                            ValidationError::InvalidTransactionSignature {
+                                id: fragment.hash().to_string(),
+                                range: self.range_check.clone(),
+                            },
+                        )));
+                    }
+                    *self.spending_counters.get_mut(&identifier).unwrap() += 1;
+                    Some(Ok(fragment))
+                }
+                _ => None,
```
Zeegomo, line 362: This actually will make the `for fragment in ....` stop if there's a fragment that's not a vote cast or tally, we should skip that instead

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -0,0 +1,648 @@
...
+pub fn deconstruct_account_transaction<P: chain_impl_mockchain::transaction::Payload>(
+    transaction: &TransactionSlice<P>,
+) -> (P, account::Identifier, account::Witness) {
+    let payload = transaction.payload().into_payload();
+    let account = transaction.inputs().iter().next().unwrap().to_enum();
+
+    let identifier = if let InputEnum::AccountInput(account, _) = account {
+        account.to_single_account().unwrap()
+    } else {
+        panic!("cannot handle utxo inputs");
```
mzabaluev, line 188: Unless this kind of thing is screened in `valid_vote_cast`, someone could break our tally tool by submitting a fragment with UTxO inputs or witnesses (that don't have to be valid). We should plug this in a follow-up PR.

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -0,0 +1,648 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        self.fragments.next().and_then(|persistent_fragment_log| {
+            let PersistentFragmentLog { fragment, time } = persistent_fragment_log;
+            match fragment {
+                Fragment::VoteCast(ref transaction) => {
+                    let transaction_slice = transaction.as_slice();
+                    let is_valid_vote_cast = valid_vote_cast(&transaction_slice);
+                    if !is_valid_vote_cast {
+                        return Some(Err((fragment, ValidationError::InvalidVoteCast)));
+                    }
```
mzabaluev, line 314: If invalid fragments can get into the permanent log (IIRC the pre-mempool checks in the node are weaker), a single invalid fragment would break our tally. I think we should report invalid fragments and proceed.

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -0,0 +1,648 @@
...
+        &mut self,
+        identifier: &account::Identifier,
+        witness: &account::Witness,
+        transaction: &TransactionSlice<P>,
+    ) -> bool {
+        let spending_counter = self
+            .spending_counters
+            .entry(identifier.clone())
+            .or_default();
+        let (is_valid_tx, _) = verify_original_tx(
```
mzabaluev, line 278: Seems like the returned value of actual spending counter is not used anywhere.

catalyst-toolbox-lib/src/recovery/tally.rs
mzabaluev: Yes, the tool should not silently drop the rest of the chain.

mzabaluev (changes requested):
The fragment iterator returning `None` on unrecognized fragments looks like a show-stopper from integrity point of view, the rest are minor issues.

catalyst-toolbox-lib/src/recovery/tally.rs
danielSanchezQ: This error is reported later. It do not break but continues with the next one. Fragment in the error is just used for logging and attaching to unprocessed fragments report.

catalyst-toolbox-lib/src/recovery/tally.rs
danielSanchezQ: I leveraged this as an "error", mostly so the fragment will be discarded and reported with the other fragments. I think it would be good for forensics.

catalyst-toolbox-lib/src/recovery/tally.rs
mzabaluev: I see, thank you.

dkijania:
Is it possible to provide url to block0 not only the file path ?

testing/catalyst-toolbox-tests/src/tally/generator.rs
```diff
@@ -0,0 +1,292 @@
...
+
+pub struct VoteRoundGenerator {
+    block0: Block0Configuration,
+    block0_hash: Hash,
+    wallets: HashMap<Address, Wallet>,
+    committee_wallets: HashMap<Address, Wallet>,
+    voteplan_managers: HashMap<VotePlanId, VotePlanManager>,
+}
+
+fn committee_id_to_address(id: CommitteeIdDef) -> Address {
```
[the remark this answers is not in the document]
Zeegomo: I don't think this is allowed by the orphan rule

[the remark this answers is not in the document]
danielSanchezQ: We could implement it wherever CommitteeIdDef is, but not here (cannot implement traits for foreign types). Can be refactored later if needed and if that is implemented at some point.

danielSanchezQ:
@dkijania I could make it to download the block0, but since we need the logs from files anyway...an extra step of having the block0 as a file should be ok.

mzabaluev (approved):
Looks good to me. Thank you for the hard work.

testing/catalyst-toolbox-tests/src/tally/generator.rs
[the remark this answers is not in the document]
mzabaluev: Also, the trait to implement is `From`. Still, if both types are foreign, it's not possible.

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -0,0 +1,672 @@
...
+                None
+            }
+        })
+        .collect()
+}
+
+/// check that the transaction input/outputs/witnesses is valid for the ballot
+/// * Only 1 input (subsequently 1 witness), no output
+pub(super) fn valid_vote_cast(tx: &TransactionSlice<certificate::VoteCast>) -> bool {
+    if tx.inputs().nb_inputs() != 1
```
eugene-babichenko, line 127: Ugh, shall we just return the result of a boolean expression? :)

eugene-babichenko (approved):
Apart from one small note, this is looking good!

catalyst-toolbox-lib/src/recovery/tally.rs
```diff
@@ -124,14 +124,9 @@ fn voteplans_from_block0(block0: &Block) -> HashMap<VotePlanId, VotePlan> {
 /// check that the transaction input/outputs/witnesses is valid for the ballot
 /// * Only 1 input (subsequently 1 witness), no output
 pub(super) fn valid_vote_cast(tx: &TransactionSlice<certificate::VoteCast>) -> bool {
-    if tx.inputs().nb_inputs() != 1
+    tx.inputs().nb_inputs() != 1
```
eugene-babichenko, line 127: I think this should be negated since you returned `false` when the condition was true.

danielSanchezQ: good catch there!

danielSanchezQ:
Tests shown green (after some weird macos rerun).

PR #9, opened 2021-05-07
Add votes printout

Zeegomo:


dkijania:
It's working great. tested.

catalyst-toolbox-cli/src/cli/recovery/votes/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+                };
+
+                let choice = if let Payload::Public { choice } = vote_cast.payload() {
+                    choice
+                } else {
+                    panic!("cannot handle private votes");
+                };
+
+                println!(
+                    "public_key: {} | voteplan: {} | proposal index: {} |   choice: {:?} ",
```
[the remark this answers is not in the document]
Zeegomo: I think you're looking at an old version, the output can already be formatted as json, but I will add the `fragment_id` field that was missing

eugene-babichenko:
What exactly is this supposed to do, print all successfully casted votes? And what is the data source?

eugene-babichenko:
Ah ok I read the commit messages. The name was misleading, I though it is the same as the archive tool.

Zeegomo:
> What exactly is this supposed to do, print all successfully casted votes? And what is the data source?

It should print all votes recorded in persistent fragment logs and the ones that successfully validated mostly for debugging purposes and checking sentry data

catalyst-toolbox-cli/src/cli/recovery/votes.rs
```diff
@@ -0,0 +1,117 @@
...
+                })
+                .unzip();
+
+        let non_filtered_votes = group_by_voter(original);
+
+        let filtered_fragments = VoteFragmentFilter::new(block0, 0..1000, to_filter.into_iter())
+            .unwrap()
+            .filter_map(Result::ok)
+            .map(|(f, sc)| (f, Some(sc)))
+            .collect::<Vec<_>>();
```
danielSanchezQ, line 102: Nit: I think we went for `filtered_fragments: Vec<_>` instead of doing that in the collect. Just team wise code style.

catalyst-toolbox-cli/src/cli/recovery/votes.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        let filtered_votes = group_by_voter(filtered_fragments);
+
+        let res = serde_json::json!({
+            "original": non_filtered_votes,
+            "filtered": filtered_votes,
+        });
```
danielSanchezQ, line 109: Nit: We could have a struct with serialize/deserialize for this (as it could be helpful at some point for parsing the log).

danielSanchezQ (commented):
LGTM. Just left a few optional ideas there.

mzabaluev (approved):

catalyst-toolbox-cli/src/cli/recovery/votes.rs
mzabaluev: Wasn't the recommendation to use whatever is more natural in the context? But yeah, here perhaps the type annotation would be better than turbofish.

PR #13, opened 2021-05-20
Replay plain transactions

Zeegomo:
Add support for replaying plain transactions, that in theory could be used as a primitive form of vote power delegation, but are not used by the app itself.

Allowed transactions are restricted to single output and account input, in practice it means vote participants can pass around the stake among themselves, (or to an account / utxo not in block 0, but they won't be able to reclaim the stake back or vote from that wallet).

Also adds a check for balanced transactions

### DIFF
--- a/src/recovery/tally.rs
+++ b/src/recovery/tally.rs
@@ -77,6 +77,9 @@ pub enum Error {
 
     #[error("Multiple outputs for a single transaction are not supported")]
     UnsupportedMultipleOutputs,
+
+    #[error("Account {0} is not known")]
+    AccountNotFound(String),
 }
 
 fn timestamp_to_system_time(ts: SecondsSinceUnixEpoch) -> SystemTime {
@@ -503,6 +506,7 @@ pub fn recover_ledger_from_logs(
 
 struct FragmentReplayer {
     wallets: HashMap<Address, Wallet>,
+    non_voting_wallets: HashMap<Address, Wallet>,
     voteplans: HashMap<VotePlanId, VotePlan>,
     old_block0_hash: Hash,
     fees: LinearFee,
@@ -554,6 +558,7 @@ impl FragmentReplayer {
         Ok((
             Self {
                 wallets,
+                non_voting_wallets: HashMap::new(),
                 voteplans,
                 old_block0_hash: block0.header.id().into(),
                 fees,
@@ -563,7 +568,7 @@ impl FragmentReplayer {
     }
 
     // rebuild a fragment to be used in the new ledger configuration with the account mirror account.
-    fn replay(&mut self, fragment: Fragment) -> Result<Fragment, Error> {
+    fn replay(&mut self, fragment: Fragment) -> Result<(Fragment, &mut Wallet), Error> {
         match fragment {
             Fragment::VoteCast(ref transaction) => {
                 let transaction_slice = transaction.as_slice();
@@ -610,8 +615,7 @@ impl FragmentReplayer {
                         &choice,
                     )
                     .unwrap();
-                wallet.confirm_transaction();
-                Ok(res)
+                Ok((res, wallet))
             }
             Fragment::Transaction(ref transaction) => {
                 let transaction_slice = transaction.as_slice();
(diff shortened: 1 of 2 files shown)


mzabaluev (approved):
Looks good, but I see that the extension is more specific than I expected.
Still, it covers more possible transactions.

src/recovery/tally.rs
```diff
@@ -548,54 +543,89 @@ impl FragmentReplayer {
...
+                let output = transaction_slice.outputs().iter().next().unwrap();
+                let output_address = self
+                    .wallets
+                    .get(&output.address.into())
+                    .ok_or_else(|| Error::NonVotingAccount(address.to_string()))?
+                    .address();
+                let wallet = self
+                    .wallets
+                    .get_mut(&address)
+                    .ok_or_else(|| Error::NonVotingAccount(address.to_string()))?;
```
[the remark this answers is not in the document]
Zeegomo: For that to happen we need to allow the creation of new account wallets as well, which should not actually be too hard. 
Those new accounts should not be allowed to vote though as per the requirements since they are not in block 0

src/recovery/tally.rs
```diff
@@ -476,9 +479,25 @@ pub fn recover_ledger_from_logs(
...
         };
-        if let Some(new_fragment) = new_fragment {
-            ledger = ledger.apply_fragment(&ledger.get_ledger_parameters(), &new_fragment, block_date)
-                .expect("Should be impossible to fail, since we should be using proper spending counters and signatures");
+        if let Some((new_fragment, wlt)) = new_fragment {
+            let new_ledger =
+                ledger.apply_fragment(&ledger.get_ledger_parameters(), &new_fragment, block_date);
+            match new_ledger {
+                Ok(new_ledger) => {
+                    if let Some(wlt) = wlt {
```
danielSanchezQ, line 487: What is `wlt`?

Zeegomo: Short for `wallet`, let me rename it so it's easier to read. I used it in some places to avoid wrapping the line, but here is not necessary

src/recovery/tally.rs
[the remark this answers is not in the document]
eugene-babichenko: > I think it's valid on the blockchain and may have legitimate uses

But is this practical in the context of Catalyst? I mean, yes, it has its use cases, but we won't be able to distribute rewards to such accounts so there is little to no incentive to do so.

src/recovery/tally.rs
```diff
@@ -267,29 +279,51 @@ impl<I: Iterator<Item = PersistentFragmentLog>> VoteFragmentFilter<I> {
...
+        );
+
+        if !valid {
+            return Err(ValidationError::InvalidTransactionSignature {
+                id: fragment_id.to_string(),
+                range: self.range_check.clone(),
+            });
+        }
+
+        // check if fragment was processed already
```
eugene-babichenko, line 318: I think it would make sense to check this before doing all other checks. Even before deconstructing a transaction.

src/recovery/tally.rs
```diff
@@ -476,9 +479,25 @@ pub fn recover_ledger_from_logs(
...
+                    ledger = new_ledger;
+                }
+                // Checking the account balance when validating the fragment would require to keep
+                // track of movements as well (i.e. re-implement a ledger)
+                Err(ledger::Error::Account(LedgerError::ValueError(
+                    ValueError::NegativeAmount,
+                ))) => {
+                    warn!("Account balance not enough to process fragment")
+                }
+                Err(_) => panic!("Should be impossible to fail, since we should be using proper spending counters and signatures"),
```
eugene-babichenko, line 499: `unreachable!` is more suitable here.

src/recovery/tally.rs
```diff
@@ -290,6 +290,13 @@ impl<I: Iterator<Item = PersistentFragmentLog>> VoteFragmentFilter<I> {
         transaction: &TransactionSlice<P>,
         fragment_id: FragmentId,
     ) -> Result<SpendingCounter, ValidationError> {
+        // check if fragment was processed already
+        if self.replay_protection.contains(&fragment_id) {
+            return Err(ValidationError::DuplicatedFragment { id: fragment_id });
+        }
+
+        self.replay_protection.insert(fragment_id);
```
eugene-babichenko, line 298: Is early insert OK though? This still may fail, but you already mutated the state.

Zeegomo: Mmh, other parts of fragment validation are actually stateless (apart from spending counter, but it's not really that important to get it precisely and it's not updated anyway if the transaction is not correct).

Now that I look at it, we allow replay of votes if the initial submission was not inside the voting phase (same for tally), and in general I think you're right, the (real) ledger behavior would be to ignore completely invalid transactions

src/recovery/tally.rs
mzabaluev: > to avoid wrapping the line

Plug in rustfmt as the autoformatter already :)

src/recovery/tally.rs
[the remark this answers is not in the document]
mzabaluev: What if they vote for change, not for personal reward? :smile: 

Yeah, it's more a wish to cover more of the lower hanging fruits here. It's OK to just report any of the unexpected transactions, to deal with them later if somebody submits them.

mzabaluev:
Please rebase the branch to resolve the merge conflicts, and let's wrap it up with what is already supported. Technically it's approved already, but let's have another look after the rebase.

src/recovery/tally.rs
Zeegomo: I do, I just like to avoid having multiple lines with parenthesis that feels in the wrong place like:
```
// rustfmt produced smth like this
long line if
{
  stuff
}
```
if this can be avoided by a few characters

mzabaluev (approved):

PR #18, opened 2021-05-26
Sentry logs utilities

danielSanchezQ:
This PR contains utilities to request, load and parse sentry logs.
It also has some functions to extract common log data from sentry and persistent fragments logs and being able to compare them.


dkijania:
So far looks very good. Do you plan to expose above code as cli ?

danielSanchezQ:
> 
> 
> So far looks very good. Do you plan to expose above code as cli ?

I think we could merge this and I can work on a CLI tool to report the differences between sentry logs and permanent fragment ones. Would that be enough @dkijania ?

dkijania:
Sure. It would also be nice if i could have option to just download sentry logs without comparison. For example during fund i would like to check if everything is ok just limited number of sentry logs

dkijania (approved):

PR #21, opened 2021-06-02
Logs cli utilities

danielSanchezQ:


src/bin/cli/logs/compare.rs
```diff
@@ -0,0 +1,150 @@
...
+            load_logs_from_file(path)?
+        } else {
+            unreachable!()
+        };
+
+        let sentry_logs_data: Vec<SentryFragmentLog> = sentry_logs
+            .iter()
+            .enumerate()
+            .filter_map(
+                |(i, raw_log)| match raw_log.get("message").map(|v| v.as_str()).flatten() {
```
Zeegomo, line 71: `flat_map` in place of `map(..).flatten()`?

src/bin/cli/logs/compare.rs
```diff
@@ -0,0 +1,150 @@
...
+    },
+}
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct Compare {
+    #[structopt(long, conflicts_with = "sentry-logs")]
+    sentry_url: Option<Url>,
+
+    #[structopt(long, conflicts_with = "sentry-logs")]
```
Zeegomo, line 37: Could this conflict be handled with an enum and a `structopt(flatten)`?

danielSanchezQ: IIRC `conflicts_with` didn't work well with flatten attributes.  It happen before. But I'll double check.

src/bin/cli/logs/compare.rs
danielSanchezQ: It is an `Option`. I could flat_map earlier but I would have to compose a larger piece I think.

Zeegomo: yeah sorry, I meant `and_then` (just a different name for a flatmap on `Option`/`Result`)

src/bin/cli/logs/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    #[error(transparent)]
+    CompareError(#[from] compare::Error),
+}
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub enum Logs {
+    Sentry(sentry::SentryLogs),
+    Compare(compare::Compare),
```
mzabaluev, line 26: Please add documentation / help strings to explain what the commands do.

danielSanchezQ: Absolutely

src/bin/cli/logs/compare.rs
danielSanchezQ: I removed this option, it is better to use downloaded logs than directly use them, since it may be a bit slow.

danielSanchezQ:
Supersede by #26

PR #26, opened 2021-06-14
Sentry stats lib + cli commands

danielSanchezQ:
This change works on top of #21 
It adds the stats command to the cli and the utilities for it.

### DIFF
--- a/src/bin/cli/logs/sentry/download.rs
+++ b/src/bin/cli/logs/sentry/download.rs
@@ -18,10 +18,12 @@ pub enum Mode {
 
 #[derive(StructOpt)]
 pub struct DateFilter {
+    // Optional [From, To] date range, start
     #[structopt(long, parse(try_from_str = DateTime::parse_from_rfc3339))]
-    start_date: Option<DateTime<FixedOffset>>,
+    from: Option<DateTime<FixedOffset>>,
+    // Optional [From, To] date range, end
     #[structopt(long, parse( try_from_str = DateTime::parse_from_rfc3339))]
-    end_date: Option<DateTime<FixedOffset>>,
+    to: Option<DateTime<FixedOffset>>,
 }
 
 #[derive(StructOpt)]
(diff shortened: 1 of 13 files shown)


src/bin/cli/logs/sentry/download.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+fn filter_logs_by_date(
+    logs: impl Iterator<Item = RawLog>,
+    dates: DateFilter,
+) -> impl Iterator<Item = RawLog> {
+    let DateFilter { from, to } = dates;
+
+    logs.take_while(move |l| {
```
Zeegomo, line 93: Does this work though? I think `take_while` stops after the first `false` value, so if the first log is before `from_date` no logs will be returned even though there are valid ones in the specified window

src/logs/sentry.rs
```diff
@@ -94,119 +104,183 @@ impl IntoIterator for LazySentryLogs {
...
+}
+
+pub struct RegexMatch {
+    key: String,
+    re: Regex,
+    total_checked: usize,
+    pub matches: usize,
+}
+
+pub enum SentryLogsStatChecker {
```
Zeegomo, line 132: I think trait objects could be used without the need to have an enum, but that's just personal preference

src/logs/sentry.rs
```diff
@@ -94,119 +104,183 @@ impl IntoIterator for LazySentryLogs {
...
-            choice: log.choice,
-            fragment_id: log.fragment_id,
+impl Stat for SuccessfulScan {
+    fn check_raw_log(&mut self, log: &RawLog) {
+        if raw_log_message_starts_with(log, REGISTERED_MESSAGE) {
+            self.total += 1;
         }
     }
+
+    fn report(&self) {
```
Zeegomo, line 164: Maybe pass a buffer or similar like `std::fmt::Formatter` to be a little more flexible if we don't want to pollute stdout?

src/bin/cli/logs/sentry/download.rs
danielSanchezQ: Uhm, what do you mean? logs come sorted by date. `from_date` is the earliest one of them.

src/logs/sentry.rs
danielSanchezQ: Aah, yeah, that is indeed better though. Will do.

src/logs/sentry.rs
danielSanchezQ: Yeah, but the idea behind this was to be able to access some public internal attributes if needed. Which you can't do with traits directly iirc.

src/bin/cli/logs/sentry/download.rs
Zeegomo: Ah sorry, I thought they were sorted with increasing dates, but they are reversed

src/logs/sentry.rs
```diff
@@ -161,8 +163,8 @@ impl Stat for SuccessfulScan {
         }
     }
 
-    fn report(&self) {
-        println!("Total successful scans: {}", self.total);
+    fn report(&self, formatter: &mut std::fmt::Formatter) -> fmt::Result {
+        formatter.write_fmt(format_args!("Total successful scans: {}", self.total))
```
Zeegomo, line 167: Why not the `write/writeln` macro?

danielSanchezQ: Just used to do it like that, but totally true.

dkijania (approved):

dkijania:
Looking good. thanks!

PR #27, opened 2021-06-18
added basic notification test

dkijania:
it requires two env variables to be set in order to hide api keys


env.bat
```diff
@@ -0,0 +1,2 @@
+SET NOTIFICATION_APP_CODE=60A7F-5F923
+SET NOTIFICATION_ACCESS_TOKEN=D8Ro3D5KxjjTIviNbTpKdb2oq6amw4Xq19HhnP0itHCFaQiI9NlDlyMJp2cPtBkYUeij9UvY4puqQMDe7W0L
```
danielSanchezQ, line 2: I think we do not need this anymore. Everything is setup through gh secrets.

danielSanchezQ (approved):
Tests green, including new. 
LGTM

env.bat
dkijania: Oh, forgot to ignore this file :(

PR #30, opened 2021-07-07
Ideascale importer

danielSanchezQ:
Use ideascale API to automate creation of json files with the content necessary for the initialization of VSS and voteplan proposal.

This PR supersede https://github.com/input-output-hk/vit-servicing-station/pull/172 (which will be deprecated).

### DIFF
--- a/src/bin/cli/ideascale/mod.rs
+++ b/src/bin/cli/ideascale/mod.rs
@@ -5,7 +5,15 @@ use catalyst_toolbox::ideascale::{
 use structopt::StructOpt;
 
 use serde::Serialize;
+use std::fmt::{Display, Formatter};
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
+
+#[derive(Debug)]
+pub enum VoteType {
+    Public,
+    Private,
+}
 
 #[derive(Debug, StructOpt)]
 pub enum Idescale {
@@ -23,9 +31,14 @@ pub struct Import {
     #[structopt(long)]
     api_token: String,
 
+    #[structopt(long)]
     threshold: i64,
+
+    #[structopt(long)]
     rewards_info: String,
-    chain_vote_type: String,
+
+    #[structopt(long)]
+    chain_vote_type: VoteType,
 
     /// Path to a json file containing the related rewards per challenge.
     /// `{ challenge_id:  reward_amount}`
@@ -56,7 +69,10 @@ impl Import {
             save_folder,
         } = self;
 
-        let runtime = tokio::runtime::Builder::new_multi_thread().build()?;
+        let runtime = tokio::runtime::Builder::new_multi_thread()
+            .enable_io()
+            .enable_time()
+            .build()?;
 
         let idescale_data =
             futures::executor::block_on(runtime.spawn(fetch_all(*fund, api_token.clone())))??;
@@ -66,7 +82,7 @@ impl Import {
 
         let funds = build_fund(&idescale_data, *threshold, rewards_info.clone());
         let challenges = build_challenges(&idescale_data, &rewards);
-        let proposals = build_proposals(&idescale_data, chain_vote_type, *fund);
+        let proposals = build_proposals(&idescale_data, &chain_vote_type.to_string(), *fund);
 
         dump_content_to_file(
             funds,
@@ -90,6 +106,28 @@ impl Import {
         )?;
 
         Ok(())
+    }
+}
+
+impl FromStr for VoteType {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "public" => Ok(Self::Public),
+            "private" => Ok(Self::Private),
+            _ => Err("Only 'public' or 'private' are correct values for VoteType"),
+        }
+    }
+}
+
+impl Display for VoteType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let s = match self {
+            VoteType::Public => "public",
+            VoteType::Private => "private",
+        };
+        write!(f, "{}", s)
     }
 }
 
--- a/src/bin/cli/ideascale/mod.rs
+++ b/src/bin/cli/ideascale/mod.rs
@@ -35,10 +35,6 @@ pub struct Import {
     #[structopt(long)]
     threshold: i64,
 
-    /// Fund rewards description
-    #[structopt(long)]
-    rewards_info: String,
-
     /// either "public" or "private"
     #[structopt(long)]
     chain_vote_type: VoteType,
@@ -62,7 +58,6 @@ impl Import {
             fund,
             api_token,
             threshold,
-            rewards_info,
             chain_vote_type,
             save_folder,
         } = self;
@@ -75,7 +70,7 @@ impl Import {
         let idescale_data =
             futures::executor::block_on(runtime.spawn(fetch_all(*fund, api_token.clone())))??;
 
-        let funds = build_fund(&idescale_data, *threshold, rewards_info.clone());
+        let funds = build_fund(&idescale_data, *threshold);
         let challenges = build_challenges(&idescale_data);
         let proposals = build_proposals(&idescale_data, &chain_vote_type.to_string(), *fund);
 
--- a/src/bin/cli/ideascale/mod.rs
+++ b/src/bin/cli/ideascale/mod.rs
@@ -32,7 +32,7 @@ pub struct Import {
     fund_goal: String,
 
     /// ideascale API token
-    #[structopt(long)]
+    #[structopt(long, env = "IDEASCALE_API_TOKEN")]
     api_token: String,
 
     /// Fund approval threshold setting
@@ -45,7 +45,7 @@ pub struct Import {
 
     /// Path to folder where fund, challenges and proposals json files will be dumped
     #[structopt(long)]
-    save_folder: PathBuf,
+    output_dir: PathBuf,
 }
 
 impl Ideascale {
@@ -64,7 +64,7 @@ impl Import {
             api_token,
             threshold,
             chain_vote_type,
-            save_folder,
+            output_dir: save_folder,
         } = self;
 
         let runtime = tokio::runtime::Builder::new_multi_thread()
--- a/src/bin/cli/ideascale/mod.rs
+++ b/src/bin/cli/ideascale/mod.rs
@@ -1,0 +1,42 @@
+use catalyst_toolbox::ideascale::{fetch_all, Error, Rewards};
+
+use structopt::StructOpt;
+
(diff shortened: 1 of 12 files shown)


src/bin/cli/ideascale/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+            save_folder,
+        } = self;
+
+        let runtime = tokio::runtime::Builder::new_multi_thread()
+            .enable_io()
+            .enable_time()
+            .build()?;
+
+        let idescale_data =
+            futures::executor::block_on(runtime.spawn(fetch_all(*fund, api_token.clone())))??;
```
eugene-babichenko, line 71: Just this maybe?

```suggestion
            runtime.block_on(fetch_all(*fund, api_token.clone()))??;
```

danielSanchezQ: Well, that is better indeed.

dkijania:
I've tested export produced by importer and has a few remarks:

**Fund**

- id: should not be random, it should be strictly defined and equal to fund number
- goal: it should be defined in separate file since it is constant, but may change in the future for example what i have in resources is:

```
[{
   "id":5,
   "goal":"Create, fund and deliver the future of Cardano.",
   "rewards_info":"Create, fund and deliver the future of Cardano.",
   "threshold":450
}]
``` 

**Challenges**

- id: as above it should be non random number strictly increasing starting from 1
- title: please remove 'F5:' (in future it could be F6 F7 and so on)
- lack of proposers_rewards field - need to consult with @MarklarIO if this is something we get from ideascale or in separate file

**Proposals**

- id: as above it should be non random number strictly increasing starting from 1
- proposal_impact_score:  it is always 0. is it something we get from ideascale ?
- missing challenge_id
- missing challenge_type
- proposal_solution is empty

src/bin/cli/ideascale/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[structopt(long)]
+    fund: usize,
+
+    /// Fund goal explanation
+    #[structopt(long)]
+    fund_goal: String,
+
+    /// ideascale API token
+    #[structopt(long)]
+    api_token: String,
```
mzabaluev, line 36: It's somewhat better security to pass the token via an environment variable rather than expose it on the command line. In CI environments this could come from a secret, be screened in logs etc.

src/bin/cli/ideascale/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[structopt(long)]
+    threshold: i64,
+
+    /// either "public" or "private"
+    #[structopt(long)]
+    chain_vote_type: VoteType,
+
+    /// Path to folder where fund, challenges and proposals json files will be dumped
+    #[structopt(long)]
+    save_folder: PathBuf,
```
mzabaluev, line 48: Let's use more traditional CLI terminology here: `--output-dir`?

src/bin/cli/ideascale/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+
+use structopt::StructOpt;
+
+use serde::Serialize;
+use std::fmt::{Display, Formatter};
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+
+#[derive(Debug)]
+pub enum VoteType {
```
mzabaluev, line 13: This could reuse `vote::PayloadTypeDef` from jormungandr-lib.

src/ideascale/fetch.rs
```diff
@@ -0,0 +1,113 @@
...
+    RequestError(#[from] reqwest::Error),
+
+    #[error("Could not get value from json, missing attribute {attribute_name}")]
+    MissingAttribute { attribute_name: &'static str },
+}
+
+#[derive(Debug, Deserialize)]
+struct Score {
+    #[serde(alias = "ideaId")]
+    id: i32,
```
mzabaluev, line 23: Not using `u32` means the idea ID could be negative?

src/ideascale/fetch.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub type Scores = HashMap<i32, f32>;
+
+static BASE_IDEASCALE_URL: Lazy<url::Url> = Lazy::new(|| {
+    "https://cardano.ideascale.com/a/rest/v1/"
+        .try_into()
+        .unwrap()
+});
+
+static CLIENT: Lazy<reqwest::Client> = Lazy::new(reqwest::Client::new);
```
mzabaluev, line 36: I'd rather use an explicit state object to instantiate the client with the API token, and make those fetch functions its methods.
Should not make much difference for the CLI tool, though.

src/ideascale/models/de.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn as_ref(&self) -> &str {
+        &self.0
+    }
+}
+
+fn deserialize_clean_string<'de, D: Deserializer<'de>>(
+    deserializer: D,
+) -> Result<String, D::Error> {
+    let mut rewards_str = String::deserialize(deserializer)?;
+    rewards_str.retain(|c| !DIRTY_CHARACTERS.contains(&c));
```
mzabaluev, line 119: `rewards_str.retain(|c| !matches!(c, '*' | '-' | '/'))` would work faster, no need for a hash set.

src/ideascale/models/de.rs
```diff
@@ -0,0 +1,152 @@
...
+    // input is not standarized, hack an early return if it is just 0 ada
+    if rewards_str.starts_with("0 ada") {
+        return Ok("0".to_string());
+    }
+    sscanf::scanf!(rewards_str.trim_end(), "${} in ada", String)
+        // trim all . or , in between numbers
+        .map(|mut s| {
+            s.retain(|c: char| c.is_numeric());
+            s
+        })
```
mzabaluev, line 148: Is it possible that some of the separators would be a decimal point? This transformation feels a bit too loose for this important amount.

src/bin/cli/ideascale/mod.rs
danielSanchezQ: Actually yes, let me let it load from both, since it is quite easy to do.

src/ideascale/fetch.rs
danielSanchezQ: No it can not, it was just most convenient that way since database already work with signed intergers so not to cast them one another.

src/ideascale/models/de.rs
danielSanchezQ: It may be, but they are cut out from input like `$600,000 in ada`, The format is the worry here. Eventually we should have a proper field that is actually a real value.

src/bin/cli/ideascale/mod.rs
danielSanchezQ: It is not exposed, iirc.

src/ideascale/fetch.rs
danielSanchezQ: Actually, those ids are not used finally except for internal linking of data, so I can fairly use unsigned ones.

src/ideascale/fetch.rs
danielSanchezQ: Yes, but it is messy to spawn clients in the async tasks later. It will avoid the api_token spreading though.

src/bin/cli/ideascale/mod.rs
danielSanchezQ: It is not exposed actually

danielSanchezQ:
@dkijania , it is ready for you to check again I think, thanks! :)


Fund

id: it wasnt random, I used the one in ideascale, it is fixed now ✅ 
goal: got this from a cli argument ✅ 

Challenges

id: same as funds, fixed too. ✅ 
title: please remove 'F5:' (in future it could be F6 F7 and so on) 
lack of proposers_rewards field - this is not needed anymore ✅ 


Proposals

id: likewise, fixed now ✅ 
proposal_impact_score: it is always 0. is it something we get from ideascale ? It was an error from we were getting them, idescale relate ✅ 
missing challenge_id -> added ✅ 
missing challenge_type -> addded ✅ 
proposal_solution is empty -> bug, fixed now ✅

src/ideascale/fetch.rs
mzabaluev: API token retrieval from an environment variable could go under the lazy initialization, but that means you can't easily have the command line option for it (which the tool probably shouldn't, as per prior discussion)

src/ideascale/models/de.rs
mzabaluev: Let's restrict this to a format that is in de-facto use: remove only commas. It would be good to validate the resulting string parses as a `u64` integer, because otherwise vit-servicing-station will fail to load.
In fact, I'd make the field type for such ADA values a newtype over `u64`.

src/ideascale/models/se.rs
```diff
@@ -0,0 +1,47 @@
+use serde::Serialize;
+
+#[derive(Debug, Serialize)]
+pub struct Challenge {
```
mzabaluev, line 4: It may become a burden to maintain two sets of structures.
Is it not possible to define `Serialize` and `Deserialize` over a single `Challenge` struct?

danielSanchezQ: I decided to use 2 because data to be deserialize and to be serialize is not the same. Not just the types but also the content itself. Some attributes are just needed to reconstruct part of the data of other structs, but may not appear in the final `de -> se` representation. Is much simpler this way instead of dealing with the serde `optional` and `skip_de/serialize` dance.

src/ideascale/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+        .await?
+        .into_iter()
+        .flatten()
+        .flatten()
+        // filter out non
+        .filter(|p| filter_proposal_by_stage_type(&p.stage_type))
+        .collect();
+
+    let mut stages: Vec<_> = fetch::get_stages(api_token.clone()).await?;
+    stages.retain(|stage| filter_stages(stage, &funnels));
```
danielSanchezQ, line 67: @mzabaluev Proposed solution for scores retrieval.

src/ideascale/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+        })
+        .collect()
+}
+
+fn filter_proposal_by_stage_type(stage: &str) -> bool {
+    matches!(stage, "Governance phase" | "Assess QA")
+}
+
+fn filter_stages(stage: &Stage, stage_label: &str, funnel_ids: &HashMap<u32, Funnel>) -> bool {
+    stage.label.to_ascii_lowercase() == stage_label.to_ascii_lowercase()
```
mzabaluev, line 207: This allocates a string from `stage_label` every time that could be pre-normalized once instead, in the code calling `filter_stages`.

mzabaluev (approved):
Looks good. Test failures seem to be unrelated.

PR #39, opened 2021-08-09
Reviews import

danielSanchezQ:
Simple command that takes the vCA aggregated file and exports a csv/json file with the data ready for being imported into the VitSS DB.


dkijania:
I think a single test on example vca and assertion on output will do. I will take care of automated e2e tests in vit-testing project

danielSanchezQ:
@dkijania I added a simple test checking number of lines for output.

Zeegomo (commented):
What about using csv instead of the proprietary xlsx format? I see no advantage in having that, especially since we only use a single file anyway, at least in my understanding

danielSanchezQ:
> What about using csv instead of the proprietary xlsx format? I see no advantage in having that, especially since we only use a single file anyway, at least in my understanding

@Zeegomo you are right. At the beginning was because there were a chance of the need to read 2 sheets. But finally we just needed one. I will change that.

danielSanchezQ:
Superseded by #58

PR #44, opened 2021-08-13
Document QR code protocol and qr-code command

mzabaluev:
Add documentation on the private key encryption protocol
and document the qr-code command in README.

### DIFF
--- a/doc/private-key-and-qr-code.md
+++ b/doc/private-key-and-qr-code.md
@@ -1,0 +1,58 @@
+# Private Key Transfer Protocol
+
+The encoding described in this document is used to export a private key
+for use in another application.
+
+In Catalyst, this format is used in the QR code used to pass the private key
+from a wallet to the mobile application.
+
+## Serialization format
+
+We will expect the scheme to be updateable. So the first byte of the data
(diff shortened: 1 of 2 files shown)


doc/private-key-and-qr-code.md
```diff
@@ -0,0 +1,58 @@
...
+   * 12983 iterations
+   * Use the SALT
+4. Encrypt the data (or decrypt)
+   * Use ChaCha20Poly1305
+   * Use the symmetric encryption key derived in step 3
+   * Use the NONCE
+
+Outputs:
+* Encode the result in the format defined in the previous section
+* Memzero everything else
```
Zeegomo, line 58: What do you mean by "memzero everything else"? If that's related to the symmetric key used during the process I wouldn't call it part of the output

Zeegomo (approved):

doc/private-key-and-qr-code.md
mzabaluev: I copied and pasted from an earlier doc. This looks like a meaningless addition: the encoding does not leave a room for anything "else", any framing that would contain this encoding (which BTW does not have an inherently defined length if we accept multiple keys in the encoded data field) is outside of this specification.

PR #46, opened 2021-08-16
Allow only one vote per account/proposal - Fix broken tests

Zeegomo:
Update tests/bin to handle cases in which an account vote for more
than one proposal.
The testing generator now produces also invalid transactions, which
should be handled correctly by the implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,6 +329,7 @@ dependencies = [
  "futures",
  "hex",
  "image",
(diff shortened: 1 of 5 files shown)


danielSanchezQ (approved):

Zeegomo:
Duplicate of https://github.com/input-output-hk/catalyst-toolbox/pull/47

PR #48, opened 2021-08-16
Move secrets to job level

Zeegomo:


mzabaluev (approved):
I don't understand why the step method does not work, but it's also cleaner this way.

I have checked that the secrets do not leak into the logs.

PR #49, opened 2021-08-16
Vote check tool

Zeegomo:
Add tool to check votes (both private and public) have been correctly tallied using node storage.
Vote verification follows this plan:

    Start a new node with the storage containing the full blockchain history to validate that all ledger operations.
    Check that the election results obtained are the same as provided
    Check that the transactions containing your votes were indeed included in a block in the main chain

Requires Jormungandr to be installed in the system

### DIFF
--- a/src/vote_check/mod.rs
+++ b/src/vote_check/mod.rs
@@ -1,0 +1,128 @@
+mod explorer;
+
+use assert_fs::{fixture::PathChild, TempDir};
+use explorer::{transaction_by_id, TransactionById};
+use graphql_client::{GraphQLQuery, Response};
+use jormungandr_integration_tests::common::jormungandr::{
+    JormungandrError, JormungandrProcess, Starter, StartupError, StartupVerificationMode,
+};
+use jormungandr_lib::interfaces::{Log, LogEntry, LogOutput, VotePlanStatus};
+use jormungandr_testing_utils::testing::{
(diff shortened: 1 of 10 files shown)


src/vote_check/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+    node::{ExplorerError, RestError},
+    Block0ConfigurationBuilder, JormungandrParams, NodeConfigBuilder,
+};
+use std::path::PathBuf;
+use std::time::Duration;
+
+const JORMUNGANDR_APP: &str = "jormungandr";
+const JORMUNGANDR_LOG_FILE: &str = "node.log";
+const JORMUNGANDR_CONFIG_FILE: &str = "node_config.yaml";
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(30);
```
danielSanchezQ, line 20: Could we have some configuration struct with defaults (for this values)?.
It could be loaded from the CLI as an optional config file path.

Zeegomo: apart from the jormungandr binary, which could in fact be exposed as a cli parameter, the other are really just used internally (they are the names of the file it will create inside the temp dir). I don't think it's worth the effort as I doubt anybody will ever have to deal with them in this setting.

danielSanchezQ (approved):
Needs rebasing, otherwise looks ok :)

PR #51, opened 2021-08-19
Private vote

Zeegomo:
Add support for private votes in the recovery tool and update relevant tests

### DIFF
--- a/src/recovery/tally.rs
+++ b/src/recovery/tally.rs
@@ -18,9 +18,10 @@ use chain_impl_mockchain::{
     certificate,
     fragment::Fragment,
     ledger::{self, Ledger},
+    testing::scenario::FragmentFactory,
     transaction::InputEnum,
     value::ValueError,
-    vote::{CommitteeId, Payload, VoteError, VotePlanLedgerError},
+    vote::{CommitteeId, VoteError, VotePlanLedgerError},
 };
 use chain_time::{Epoch, Slot, SlotDuration, TimeEra, TimeFrame, Timeline};
 use imhamt::UpdateError;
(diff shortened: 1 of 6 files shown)


src/recovery/tally.rs
```diff
@@ -18,9 +18,10 @@ use chain_impl_mockchain::{
     certificate,
     fragment::Fragment,
     ledger::{self, Ledger},
+    testing::scenario::FragmentFactory,
```
mzabaluev, line 21: Curious that we need to depend on things from the testing API, from the looks of it.

mzabaluev: Also, these use statements look messy, there's overlap with paths used above.

src/recovery/tally.rs
```diff
@@ -577,46 +584,21 @@ impl FragmentReplayer {
...
-                let res = wallet
-                    .issue_vote_cast_cert(
-                        &self.old_block0_hash,
-                        &self.fees,
-                        vote_plan,
-                        proposals_idx,
-                        choice,
-                    )
-                    .unwrap();
+                let res = FragmentFactory::new(self.old_block0_hash.into_hash(), self.fees)
```
mzabaluev, line 599: I'd say we should move this to the core API, or maybe improve the regular tx builder so that it's as convenient to use as the testing stuff.

Zeegomo: Part of the reasons it's more convenient to use this is that it relies on other testing stuff like `Wallet`, which we were using previously anyway. AFAIK we don't have the concept of a wallet in jormungandr production code, thus making it more difficult to streamline such operations and/or "promote" `FragmentFactory` to the core API.

Maybe we could see if chain-wallet-libs can offer a a valid option for a production ready wallet which is also easy enough to work with.

PR #57, opened 2021-09-20
Split kedqr command into hash and img

dkijania:
This pr is an attempt to move all qr code related code done by @enzoc4 in jormungandr-testing-utils project to this one. This required small rewrite:
- split hash and image generation
- add separate command for each


src/bin/cli/kedqr/hash.rs
```diff
@@ -0,0 +1,70 @@
...
+        // process output
+        match output {
+            Some(path) => {
+                // save qr code to file, or print to stdout if it fails
+                let mut file = File::create(path)?;
+                file.write_all(qr.as_bytes())?;
+            }
+            None => {
+                // prints qr code to stdout when no path is specified
+                println!();
```
enzoc4, line 64: Why is this blank line needed?

And in any case, I think I prefer `println!("\n{}", qr);`

src/bin/cli/kedqr/hash.rs
```diff
@@ -0,0 +1,70 @@
...
+};
+use structopt::StructOpt;
+
+/// QCode CLI toolkit
+#[derive(Debug, PartialEq, StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct QrCodeHashCmd {
+    /// Path to file containing ed25519extended bech32 value.
+    #[structopt(short, long, parse(from_os_str))]
+    input: PathBuf,
```
enzoc4, line 21: these arguments seems to be the same than for the `QrCodeHashImg`, should we use [this](https://docs.rs/structopt/0.3.23/structopt/#flattening) to avoid duplication?

src/bin/cli/kedqr/img.rs
```diff
@@ -0,0 +1,71 @@
...
+            .create(false)
+            .read(true)
+            .write(false)
+            .append(false)
+            .open(&input)
+            .expect("Could not open input file.");
+
+        let mut reader = BufReader::new(key_file);
+        let mut key_str = String::new();
+        let _key_len = reader
```
enzoc4, line 41: is there any reason for this `let` binding? although I think it's fine anyway.

dkijania:
I've reimplemented e2e test to rely on code rather than static files, which can be obsolete (like block0.bin) after changes in jormungandr

danielSanchezQ:
There is code from the tally recovery in this PR. It is a bit weird. Where did you branched it from?

dkijania:
Oh yes, good catch. my goal is to merge it to catalyst-fund6 instead of main.

danielSanchezQ (approved):
Lgtm

Zeegomo:
Do we need to port it to master as well?

danielSanchezQ:
@Zeegomo yes, we would need to. Which makes me wonder what was that in the tally recovery that is not in master 🤔

Zeegomo:
I think it was some refactor of the `Replay` struct to be able to use it from other projects (commit https://github.com/input-output-hk/catalyst-toolbox/pull/57/commits/693521144cf8a7fe451993b73263ee382314dec0) + some test changes (https://github.com/input-output-hk/catalyst-toolbox/pull/57/commits/19d0932fd278f443e5c337f7930a88640bd4d16d) . Can you confirm @dkijania ?
I didn't look too much into it, but it's merged already and it looks like it's just moving stuff around so I don't think it's worth to revert. Next time it's better to have a separate PR though

PR #58, opened 2021-09-30
Reviews import

danielSanchezQ:
Simple command that takes the vCA aggregated file and exports a csv/json file with the data ready for being imported into the VitSS DB.


danielSanchezQ:
Supersedes #39

src/vca_reviews/mod.rs
```diff
@@ -4,21 +4,14 @@ pub use tags::TagsMap;
...
 pub enum Error {
     #[error(transparent)]
-    CouldNotReadExcel(#[from] calamine::Error),
-
-    #[error(transparent)]
-    Xls(#[from] calamine::XlsxError),
-
-    #[error(transparent)]
-    CouldNotDeserialize(#[from] calamine::DeError),
+    CouldNotReadExcel(#[from] csv::Error),
```
Zeegomo, line 14: Well, it's not and excel file anymore :)

src/vca_reviews/mod.rs
```diff
@@ -4,21 +4,14 @@ pub use tags::TagsMap;
...
 
     #[error("Couldn't find workbook {0}")]
     CouldNotFindWorkbook(String),
```
Zeegomo, line 17: I think this is not needed anymore

src/vca_reviews/mod.rs
danielSanchezQ: indeed

Zeegomo (approved):

PR #62, opened 2021-10-07
Ca rewards calculation command

danielSanchezQ:


src/rewards/ca/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+    proposal_rewards_states
+        .into_iter()
+        .map(|(id, state)| {
+            let bonus_funds = approved_proposals
+                .contains(&id)
+                .then(|| bonus_proposals_rewards)
+                .unwrap_or_else(|| Funds::from(0u64));
+            let funds = match state {
+                ProposalFundsState::Unfilled(unfilled_funds) => {
+                    per_proposal_reward - unfilled_funds
```
Zeegomo, line 101: I think `underbudget_rewards` should be added to unfilled proposals as well

src/rewards/ca/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+        .zip(std::iter::repeat(Rewards::default()))
+        .collect();
+
+    for (proposal, reviews) in proposal_reviews {
+        let proposal_reward = proposal_funds.get(proposal).unwrap();
+        let proposal_rewards =
+            calculate_ca_rewards_for_proposal(proposal_reward, reviews, rewards_slots);
+
+        for (ca, rewards) in proposal_rewards {
+            *ca_rewards.get_mut(&ca).unwrap() += rewards;
```
Zeegomo, line 196: same as before, we could use `entry().or_insert(Rewards::default())`

src/rewards/ca/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+
+fn calculate_ca_rewards_for_proposal(
+    proposal_reward: &ProposalReward,
+    proposal_reviews: &[AdvisorReviewRow],
+    rewards_slots: &ProposalRewardSlots,
+) -> CaRewards {
+    let ProposalReward { state, funds } = proposal_reward;
+    let tickets_distribution = load_tickets_from_reviews(proposal_reviews, rewards_slots);
+    match state {
+        ProposalFundsState::Exact | ProposalFundsState::Unfilled(_) => {
```
Zeegomo, line 166: actually, there's no need to have separate logic, a lottery picking `filled_slots` tickets would work for all of them

src/rewards/ca/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+    Exact,
+    // Proposal has less reviews as needed so some of the funds should go back into the rewards pool
+    Unfilled(Funds),
+    // It has more reviews than fitted rewards
+    OverLoaded,
+}
+
+struct ProposalReward {
+    pub state: ProposalFundsState,
+    pub funds: Funds,
```
Zeegomo, line 34: Is it necessary to store funds per proposal? Apart from bonuses, which can be stored here, each review is worth the same irregardless of the proposal

src/bin/cli/rewards/mod.rs
```diff
@@ -17,12 +19,16 @@ pub enum Error {
 pub enum Rewards {
     /// Calculate rewards for voters base on their stake
     Voters(voters::VotersRewards),
+
+    /// Calculate community advisors rewards
+    Cas(community_advisors::CommunityAdvisors),
```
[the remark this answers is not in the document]
danielSanchezQ: Yes, didnt wanted to make it too large fot hebsake of the cli though.

src/rewards/ca/lottery.rs
```diff
@@ -0,0 +1,40 @@
...
+    tickets_to_distribute: TotalTickets,
+) -> CasWinnings {
+    let mut winnings: CasWinnings = distribution
+        .keys()
+        .cloned()
+        .zip(std::iter::repeat(0))
+        .collect();
+
+    for _ in 0..tickets_to_distribute {
+        let winner = lottery_winner(distribution);
```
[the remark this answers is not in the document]
danielSanchezQ: Aaah, yeah. I actually thought of it while implementing it, and then forgot to fix that. Thanks!

src/rewards/ca/mod.rs
danielSanchezQ: Im not sure ablut that. But it is a point of discussion because then it would have more budget again... 
But, it also makes sense to make just another iteration like that. Let me ask Marek about this.

src/rewards/ca/mod.rs
danielSanchezQ: I know about this. I made the decision so the simole implementation is straight forward, It may be a bit convoluted using a lottery just for a normal distribution.

Zeegomo: I think it's personal preference, but I would prefer having a single general implementation than multiple specific ones

src/community_advisors/models/de.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+impl AdvisorReviewRow {
+    pub fn score(&self) -> ReviewScore {
+        match (self.excellent, self.good) {
+            (true, false) => ReviewScore::Excellent,
+            (false, true) => ReviewScore::Excellent,
+            _ => {
+                // This should never happen
+                panic!(
```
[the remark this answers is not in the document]
danielSanchezQ: Uhm, actually, this brings me to the point that proposals with no reviews should actually be processed as underbudget and redistribute the rewards. Which currently we are not doing for the simple fact that they do not appear in proposals pot in the input data. I need to revisit this. Really good point @Zeegomo !

src/bin/cli/rewards/community_advisors.rs
```diff
@@ -84,6 +90,10 @@ impl CommunityAdvisors {
             &approved_proposals,
             &fund_settings.into(),
             &rewards_slots.into(),
+            Seed::from(
+                Digest::<Blake2b256>::from_str(&seed)
```
Zeegomo, line 94: This does not hash the input string, but rather tries to interpret it as a black2b256 digest. I think we should do this the other way around

src/bin/cli/rewards/mod.rs
```diff
@@ -19,6 +19,9 @@ pub enum Error {
 
     #[error("{0}")]
     InvalidInput(String),
+
+    #[error("Invalid seed: {0}")]
+    InvalidSeed(Box<dyn std::error::Error>),
```
Zeegomo, line 24: We don't need this anymore

Zeegomo (approved):
Looks good to me, let's see if we got requirements right or this needs further changes

src/community_advisors/models/de.rs
```diff
@@ -64,4 +66,29 @@ mod tests {
         let data: Vec<AdvisorReviewRow> = csv_utils::load_data_from_csv(&file_path).unwrap();
         assert_eq!(data.len(), 1);
     }
+
+    impl AdvisorReviewRow {
+        pub fn dummy(score: ReviewScore) -> Self {
```
danielSanchezQ, line 71: Is this just used for testing? If thats the case, let's move it somewhere better, as a separated function or maybe a feature flag.

Zeegomo: It's under a test module with `#[cfg(test)]` now, is that what you mean? Otherwise I'm open to suggestions, we could even implement something like proptest's `Arbitrary`.

danielSanchezQ: yeah, under `#[cfg(test)]` is ok.

src/rewards/community_advisors/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+                    rewards_slots.max_winning_tickets() % LEGACY_MAX_WINNING_TICKETS
+                );
+                total_tickets += winning_tkts
+                    * (rewards_slots.max_winning_tickets() / LEGACY_MAX_WINNING_TICKETS);
+            }
+            ProposalTickets::Fund6 { winning_tkts, .. } => total_tickets += winning_tkts,
+        }
+
+        proposals_tickets.push((id, tickets));
+    }
```
danielSanchezQ, line 77: Nitpick: we could make a function for this.

danielSanchezQ (commented):
Awesome job porting the docs examples to tests @Zeegomo

PR #67, opened 2021-10-14
Vca rewards

danielSanchezQ:
It could have been calculated directly in the excel file with a simple formula too.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,7 +99,7 @@ jobs:
         continue-on-error: false
         with:
           command: test
-          args: --locked
+          args: --locked --all-targets
 
   lints:
     name: Rust lints


src/rewards/veterans.rs
```diff
@@ -0,0 +1,35 @@
...
+    name: VeteranAdvisorId,
+    #[serde(alias = "No. of Reviews")]
+    number_of_reviews: usize,
+}
+
+pub fn calculate_veteran_advisors_rewards(
+    veteran_reviews: &[VeteranReview],
+    base_rewards: VeteranAdvisorReward,
+) -> VeteranAdvisorRewards {
+    let total_reviews: VeteranAdvisorReward = VeteranAdvisorReward::from(
```
Zeegomo, line 19: nit: type and variable names seems to disagree on what the data represents (also the type specification is redundant). I would leave this as `usize` and do the conversion later for better readability

src/rewards/veterans.rs
```diff
@@ -0,0 +1,35 @@
+use serde::Deserialize;
+
+pub type VeteranAdvisorId = String;
+pub type VeteranAdvisorReward = rust_decimal::Decimal;
+
+pub type VeteranAdvisorRewards = Vec<(VeteranAdvisorId, VeteranAdvisorReward)>;
+
+#[derive(Deserialize)]
+pub struct VeteranReview {
```
Zeegomo, line 9: nit: naming is a bit confusing. This is not actually a review, but rather holds info about a vCA

src/utils/csv.rs
```diff
@@ -0,0 +1,26 @@
...
+use serde::Serialize;
+
+use std::path::Path;
+
+pub fn load_data_from_csv<T: DeserializeOwned>(file_path: &Path) -> Result<Vec<T>, csv::Error> {
+    let mut csv_reader = csv::ReaderBuilder::new()
+        .has_headers(true)
+        .from_path(file_path)?;
+    let mut res = Vec::new();
+    for entry in csv_reader.deserialize() {
```
Zeegomo, line 11: nit: I think it should be possible to do something like `csv_reader.deserialize().collect::<Result<Vec<T>, _>>()?`

src/rewards/veterans.rs
```diff
@@ -0,0 +1,35 @@
...
+            .map(|vr| vr.number_of_reviews)
+            .sum::<usize>(),
+    );
+
+    veteran_reviews
+        .iter()
+        .map(|vr| {
+            (
+                vr.name.clone(),
+                (VeteranAdvisorReward::from(vr.number_of_reviews) / total_reviews) * base_rewards,
```
Zeegomo, line 31: It just occurred to me that this (and similar other calculations in catalyst-toolbox) returns a decimal value, whereas in the end we need to send an integer amount of lovelaces. Is the conversion done manually?

src/rewards/veterans.rs
danielSanchezQ: Indeed, let me find a more suitable naming for this.

src/utils/csv.rs
danielSanchezQ: I would love for it to, but sadly it is not iirc.

Zeegomo (approved):
We probably should add a simple test for this (and open a PR from the upstream repo branch so that CI can run)

danielSanchezQ:
@Zeegomo agree, and damn upstream branches 😆

danielSanchezQ:
migrated to #69

PR #69, opened 2021-10-15
Vca rewards

danielSanchezQ:
Using this instead of #67 for testing issues.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,7 +99,7 @@ jobs:
         continue-on-error: false
         with:
           command: test
-          args: --locked
+          args: --locked --all-targets
 
   lints:
     name: Rust lints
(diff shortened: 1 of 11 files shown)


.github/workflows/test.yml
```diff
@@ -99,7 +99,7 @@ jobs:
         continue-on-error: false
         with:
           command: test
-          args: --locked
+          args: --locked --all-targets
```
danielSanchezQ, line 102: Tests in the `bin` folder weren't being run.

resources/testing/veteran_reviews_count.csv
```diff
@@ -0,0 +1,7 @@
+name,vca_link,No. of Reviews
+Andreas,https://docs.google.com,10
```
Zeegomo, line 2: I know it's very simple but I would at least try with different n. of reviews 😂

danielSanchezQ: Well, i did it on porpouse, I wanted it to be squared 😅 , but yeah, let me fix some values.

PR #70, opened 2021-10-20
Reviews import revisited

danielSanchezQ:
Adapted reviews importer tool to new reviews schema.


danielSanchezQ:
Security audit should be cleared in an external Pr

src/vca_reviews/mod.rs
```diff
@@ -15,63 +13,27 @@ pub enum Error {
...
-    assessor: String,
-
-    #[serde(alias = "Assessment Note")]
-    note: String,
-}
-
 impl AdvisorReviewRow {
-    fn as_advisor_review(&self, tags_map: &TagsMap) -> Result<AdvisorReview, Error> {
-        Ok(AdvisorReview {
-            id: self.id as i32,
```
Mr-Leshiy, line 50: As I see this field `id` is not used, can we remove it ?
(I know that it is from the dependecy lib)

danielSanchezQ: Yeah, it is not there anymore. I think this was commented before the rebase maybe?

Mr-Leshiy: Looks it is still present https://github.com/input-output-hk/catalyst-toolbox/blob/f88382435f93e97f7bc6642a75a0b3cb4c1e7c44/src/vca_reviews/mod.rs#L19 .

danielSanchezQ: Aha, you meant from the `AdvisorReview` itself. I do not think it is possible to remove, since this structure (iirc) is used for serialization on other means (where it does actually need the id).

Mr-Leshiy (approved):
LGTM

PR #74, opened 2021-11-08
Ca rewards pipeline fixes

danielSanchezQ:
* Implemented csv merget
* Adapt proposers rewards tool for proper csv formatting (matching other tools)
* Adapt ca rewards command csv formatting too

### DIFF
--- a/scripts/python/csv_merger.py
+++ b/scripts/python/csv_merger.py
@@ -1,6 +1,7 @@
 import csv
 import glob
 from pathlib import Path
+import os
 from typing import List, Generator, Iterator, TextIO, Tuple
 import typer
 from contextlib import contextmanager
@@ -15,7 +16,7 @@ def open_files(files: Iterator[str]) -> Generator[Iterator[TextIO], None, None]:
 
 
 def search_file_pattern(pattern: str, base_path: Path) -> Iterator[str]:
-    yield from glob.iglob(pathname=pattern, root_dir=base_path)
+    yield from map(lambda file: os.path.join(base_path, file), glob.iglob(pathname=pattern, root_dir=base_path))
 
 
 def file_as_csv(file: TextIO, delimiter: chr) -> Tuple[List[str], Iterator[List[str]]]:
@@ -30,6 +31,7 @@ def merge_csv(
     input_delimiter: chr,
     output_delimiter: chr,
 ):
+    print(pattern)
     files = search_file_pattern(pattern, base_path)
     with open(output_file, "w", encoding="utf-8", newline="") as out_file:
         with open_files(files) as fs:
--- a/scripts/python/csv_merger.py
+++ b/scripts/python/csv_merger.py
@@ -16,7 +16,10 @@ def open_files(files: Iterator[str]) -> Generator[Iterator[TextIO], None, None]:
 
 
 def search_file_pattern(pattern: str, base_path: Path) -> Iterator[str]:
-    yield from map(lambda file: os.path.join(base_path, file), glob.iglob(pathname=pattern, root_dir=base_path))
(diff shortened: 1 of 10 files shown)


scripts/python/csv_merger.py
```diff
@@ -0,0 +1,62 @@
+import csv
+import glob
+from pathlib import Path
+import os
+from typing import List, Generator, Iterator, TextIO, Tuple
+import typer
```
eugene-babichenko, line 6: I am a bit wary of using external libraries in scripts. I'd rather stick with `argparse`.

danielSanchezQ: I understand the concern. But,  Imo, it shouldn't matter for the following reasons:

* We have been using it already
* It is a simple process (`pip install requirements`)
* Makes the code simpler and tidier
* At the end, we use some other deps (like `yaml`, that you will still need to install, so a few more should be ok)

scripts/python/proposers_rewards.py
```diff
@@ -510,16 +513,18 @@ def calculate_rewards(
             approval_threshold,
             total_stake_approval_threshold,
         )
-        out_stream = (
-            output_json(results)
-            if output_format == OutputFormat.JSON
-            else output_csv(results)
-        )
+
         chalenge_ouput_file_path = build_path_for_challenge(
```
Zeegomo, line 517: nit: typo

src/community_advisors/models/mod.rs
```diff
@@ -5,15 +5,16 @@ use serde::{Deserialize, Deserializer};
 pub use de::{AdvisorReviewRow, ReviewScore};
 
 pub enum ProposalStatus {
-    Approved,
-    NotApproved,
+    Funded,
```
Zeegomo, line 8: mmm, `Funded` and `Approved` are not synonyms, what's the rationale behind this?

src/utils/csv.rs
```diff
@@ -3,9 +3,12 @@ use serde::Serialize;
 
 use std::path::Path;
 
-pub fn load_data_from_csv<T: DeserializeOwned>(file_path: &Path) -> Result<Vec<T>, csv::Error> {
+pub fn load_data_from_csv<T: DeserializeOwned, const DELIMITER: u8>(
```
Zeegomo, line 6: nice const generics :)

src/community_advisors/models/mod.rs
danielSanchezQ: This was a missuse. In the original document the strings `Funded` and `NotFunded` are used, so I wanted to reflect the same here. Logically speaking, all proposals that we use in the document are **approved** but not all are **funded**.

scripts/python/csv_merger.py
eugene-babichenko: Well, I guess since we have `requirements.txt` I think it should be fine. Fair point about using other deps.

src/community_advisors/models/mod.rs
Zeegomo: Can you point me to the document? In the rewards script I'm quite sure those were supposed to be approved ones and not funded

danielSanchezQ: You can check the tags [here](https://github.com/input-output-hk/catalyst-toolbox/blob/54ced5daadf6405ef80086c49832a447057e0f3b/scripts/python/proposers_rewards.py#L21)

danielSanchezQ: @Zeegomo , You were right, we should be reading from the `meets_approval_threshold` column in the csv. It is fixed now.

Zeegomo (approved):

PR #75, opened 2021-11-25
Voters rewards ballot threshold

danielSanchezQ:


src/rewards/voters.rs
```diff
@@ -33,14 +37,65 @@ pub fn calculate_stake<'address>(
...
+    let identifier: UnspecifiedAccountIdentifier = UnspecifiedAccountIdentifier::from(buffer);
+    Ok(Address::from(chain_addr::Address(
+        discrimination,
+        Kind::Account(
+            identifier
+                .to_single_account()
+                .expect("Only single accounts are supported")
+                .into(),
+        ),
+    )))
```
danielSanchezQ, line 100: I'm a bit concern about this. I would like to strongly review it.

src/rewards/voters.rs
```diff
@@ -33,14 +37,65 @@ pub fn calculate_stake<'address>(
...
+    threshold_addresses: &AddressesVoteCount,
+    threshold: u64,
 ) -> HashMap<&'address Address, U64F64> {
     stake_per_voter
         .iter()
-        .map(|(k, v)| (*k, fixed::types::U64F64::from_num(*v) / total_stake as u128))
+        .map(|(k, v)| {
+            let reward = if *threshold_addresses
+                .get(k)
+                .unwrap_or_else(|| panic!("Missing address {} in vote count map", k))
```
Zeegomo, line 48: Would this panic if a user was in block0 but did not actually vote?

src/rewards/voters.rs
```diff
@@ -33,14 +37,65 @@ pub fn calculate_stake<'address>(
...
+                >= threshold
+            {
+                U64F64::from_num(*v) / total_stake as u128
+            } else {
+                U64F64::default()
```
Zeegomo, line 53: nit: `U64F64::ZERO` would be easier to read IMO

src/rewards/voters.rs
```diff
@@ -33,14 +37,65 @@ pub fn calculate_stake<'address>(
 pub fn calculate_reward_share<'address>(
     total_stake: u64,
     stake_per_voter: &HashMap<&'address Address, u64>,
+    threshold_addresses: &AddressesVoteCount,
+    threshold: u64,
 ) -> HashMap<&'address Address, U64F64> {
```
Zeegomo, line 42: unrelated: maybe we should use an alias for `U64F64` like we did in ca rewards

src/rewards/voters.rs
```diff
@@ -33,14 +37,65 @@ pub fn calculate_stake<'address>(
...
         .collect()
 }
 
 /// get the proportional reward from a share and total rewards amount
 pub fn reward_from_share(share: U64F64, total_reward: u64) -> fixed::types::U64F64 {
     fixed::types::U64F64::from_num(total_reward) * share
 }
+
+pub type VoteCount = HashMap<String, u64>;
+pub type AddressesVoteCount = HashMap<Address, u64>;
```
Zeegomo, line 66: What about implementing `Deserialize` on `AddressesVoteCount` not to have a type (`VoteCount`) which is actually never used in practice?

src/rewards/voters.rs
danielSanchezQ: Not sure. It depends on chain implementation now that you mention it. Maybe we could consider a defaulting 0 votes a better choice though.

src/rewards/voters.rs
danielSanchezQ: Yep, and I wanted to do that. But the hint was missing from my IDE and I took the easy path 😅

src/rewards/voters.rs
danielSanchezQ: Yep, I'll do.

src/rewards/voters.rs
danielSanchezQ: Good idea. Will do.

danielSanchezQ: Actually, I think it would be helpful to keep like this. Reason is that when debugging could be clearer that we rematch the hexlified identifiers to the Addresses themselves. It may be handier to do it in the deserialize implementation but also maybe more confusing.

src/rewards/voters.rs
eugene-babichenko: For voting we imply that the input is always of account type, so I think it's fine there.

src/bin/cli/rewards/voters.rs
```diff
@@ -22,12 +23,18 @@ pub struct VotersRewards {
     /// Reward (in LOVELACE) to be distributed
     #[structopt(long = "total-rewards")]
     total_rewards: u64,
+
+    #[structopt(long = "votes-count-path")]
```
eugene-babichenko, line 27: Nitpick: I think we don't need this if we already have `#[structopt(rename_all = "kebab-case")]`. Don't know why we've been doing this tbh.

src/rewards/voters.rs
eugene-babichenko: > Reason is that when debugging could be clearer that we rematch the hexlified identifiers to the Addresses themselves

Could be solvable with the custom `Debug` impl.

danielSanchezQ: No really, I mean in the code flow. It is an explicit call instead of some magic behind the deserialization itself.

src/bin/cli/rewards/voters.rs
danielSanchezQ: By using just the `long` attribute force you to use it as a flag parameter instead of as a positional parameter. Imho it is clearer (although I understand it is more bothersome to write the full flag) that just positional parameters.

Zeegomo: I think he meant you could simply write `#[structopt(long)]`

danielSanchezQ: Aah, well, yes. Don't know what was I thinking about though.

eugene-babichenko (approved):

PR #77, opened 2021-11-26
add double lottery to ca rewards

Zeegomo:


src/rewards/community_advisors/lottery.rs
```diff
@@ -26,13 +26,16 @@ pub fn lottery_distribution<R: Rng>(
...
         let tickets_won = std::iter::from_fn(|| {
-            indexes.next_if(|tkt| *tkt < (cumulative_ticket_index + n_tickets) as usize)
+            indexes.next_if(|tkt| *tkt < (cumulative_ticket_index + *n_tickets) as usize)
         })
         .count();
-        cumulative_ticket_index += n_tickets;
-        winnings.insert(ca, tickets_won as u64);
+        cumulative_ticket_index += *n_tickets;
+        if tickets_won > 0 {
+            winnings.insert(ca.clone(), tickets_won as u64);
```
eugene-babichenko, line 36: Conversion from `usize` to `u64` is potentially dangerous. I would not rely on the assumption of having 64-bit `usize`.

src/rewards/community_advisors/mod.rs
```diff
@@ -131,38 +142,54 @@ fn load_tickets_from_reviews(
...
+            ReviewScore::Excellent => (rev.assessor.clone(), rewards_slots.excellent_slots),
+            ReviewScore::Good => (rev.assessor.clone(), rewards_slots.good_slots),
+            _ => unreachable!("we've already filtered out other review scores"),
+        }).partition(|(_ca, tkts)| *tkts == rewards_slots.excellent_slots);
+
+    let excellent_winning_tkts = std::cmp::min(
+        excellent_tkts.len() as u64,
+        rewards_slots.max_excellent_reviews,
+    ) * rewards_slots.excellent_slots;
+    let good_winning_tkts = std::cmp::min(good_tkts.len() as u64, rewards_slots.max_good_reviews)
```
eugene-babichenko, line 158: Same here about `usize -> u64`.

src/rewards/community_advisors/lottery.rs
Zeegomo: Why? If something I would expect the opposite to be more dangerous, don't expect 128 bits archs to become that common in the near future.
Also that `usize` is actually bound to the limits of `u64` as the total number of tickets is a `u64`.

src/rewards/community_advisors/mod.rs
```diff
@@ -131,38 +142,54 @@ fn load_tickets_from_reviews(
...
-        winning_tkts: excellent_winning_tkts + good_winning_tkts,
+// Run a two stage lottery to reward community advisors
+//
+// In the first round, only excellent reviews will be taken into consideration
+// In the second, losing tickets from the first round will compete with good review
+fn double_lottery<R: Rng>(
+    stage1: TicketsDistribution,
+    mut stage2: TicketsDistribution,
+    to_distribute1: u64,
+    to_distribute2: u64,
```
danielSanchezQ, line 177: nitpick: I do not clearly see the meaning on `to_distribute1/2`. Maybe `distribute_first_round`, `distribute_second_round`?

src/rewards/community_advisors/lottery.rs
eugene-babichenko: I'm talking about the possibility of running it on a 32-bit system :)

eugene-babichenko: Ah well, I was dumb, it would break with the opposite conversion. Sorry.

danielSanchezQ (approved):

PR #80, opened 2021-12-14
adjustment for ideascale import after tests

dkijania:
Fixed some minor stuff when testing ideascale import:

- added white-space when querying for particular fund
- renamed `proposal goal` mapping
- made rewards deserializer more universal since it can hold different crypto from ada
- clippy issues

### DIFF
--- a/src/ideascale/mod.rs
+++ b/src/ideascale/mod.rs
@@ -85,7 +85,7 @@ pub async fn fetch_all(
         funnels,
         fund: funds
             .into_iter()
-            .find(|f| f.name.as_ref().contains(&format!("Fund{}", fund)))
+            .find(|f| f.name.as_ref().contains(&format!("Fund {}", fund)))
             .unwrap_or_else(|| panic!("Selected fund {}, wasn't among the available funds", fund)),
         challenges: challenges.into_iter().map(|c| (c.id, c)).collect(),
         proposals: proposals.map(|p| (p.proposal_id, p)).collect(),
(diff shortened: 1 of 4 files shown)


danielSanchezQ (approved):

src/ideascale/mod.rs
```diff
@@ -85,7 +85,7 @@ pub async fn fetch_all(
         funnels,
         fund: funds
             .into_iter()
-            .find(|f| f.name.as_ref().contains(&format!("Fund{}", fund)))
+            .find(|f| f.name.as_ref().contains(&format!("Fund {}", fund)))
```
danielSanchezQ, line 88: Why oh why?

src/ideascale/models/custom_fields.rs
```diff
@@ -21,7 +21,7 @@ impl Default for CustomFieldTags {
             proposal_solution: "problem_solution".to_string(),
             proposal_brief: "challenge_brief".to_string(),
             proposal_importance: "importance".to_string(),
-            proposal_goal: "describe_your_solution_to_the_problem".to_string(),
+            proposal_goal: "how_does_success_look_like_".to_string(),
```
danielSanchezQ, line 24: Again, this should bump on tests but we should have a ticket from whoever changed or decide to change this.

danielSanchezQ: Btw, this is part of configuration input for the command. It could be done externally if something changes, no need to change it code wise.

src/ideascale/mod.rs
dkijania: they added space in name :) But agree i could do regex

src/ideascale/models/custom_fields.rs
dkijania: Agree, but in code is much easier than in config file

Zeegomo: I still think that if we can update the default with the current value that is a nice UX improvement

Zeegomo (approved):

PR #82, opened 2021-12-16
Voters rewards addresses knowledge

danielSanchezQ:
I am opening this draft as part of documentation.

We needed to get the addresses (main cardano) of the voters. Currently we just got the addresses used by jormungandr when voting (pk). And we needed something to match them in order to fill the gap.

The main idea is to query and filter those from the `db-sync`. I have already successfully(?) query those. But we need to filter them by actual proper transactions and then get the ones with the highest nonce per voter.
For doing that my was to take two steps:

1. Query everything to the db with a python script and dump the relevant information into a json file.
2. Take that json as input in the tool so it can be processed.

I could be done entirely by the tool (including query), but my attempts to use something as simple as `sqlx` were not pleasant. So I kept the original python experimental script.


scripts/python/db_sync_transactions_dump.py
```diff
@@ -0,0 +1,56 @@
...
+sig_table AS (
+    select tx_id, json AS signature from tx_metadata where key = '61285'
+    )
+SELECT tx.hash, tx_id, metadata, signature, block_index
+FROM meta_table
+    INNER JOIN tx ON tx.id = meta_table.tx_id
+    INNER JOIN sig_table USING(tx_id)
+    INNER JOIN block ON block.id = tx.block_id
+ORDER BY metadata -> '4' ASC;
+"""
```
danielSanchezQ, line 21: This comes from a [conversation ](https://input-output-rnd.slack.com/archives/CC0CLNEH1/p1635432012025000) with Samuel. 
Is it cryptic? yeah, it is.

src/rewards/voters/addresses.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub nonce: u64,
+}
+
+#[derive(Deserialize)]
+pub struct Transaction {
+    hash: String,
+    tx_id: String,
+    metadata: Metadata,
+    signature: String,
+    block_index: u32,
```
danielSanchezQ, line 19: The idea is to match the output of the python script, load it with `serde_json` and process it afterwards.

src/rewards/voters/addresses.rs
```diff
@@ -0,0 +1,46 @@
...
+    // Next, we need to hash the cbor data. We input the cbor data to Blake2b256 and verify
+    // the signature over that message
+    let mess_bytes = hex::decode(&"a119ef64a40158200036ef3e1f0d3f5989e2d155ea54bdb2a72c4c456ccb959af4c94868f473f5a002582086870efc99c453a873a16492ce87738ec79a0ebd064379a62e2c9cf4e119219e035839009f0186a15f50592b311b57980e06cf9e791dfcb998a1fb8bfd65d06eae3a0a7aeda4aea522e74e4fe36759fca80789a613a58a4364f6ecef041904d2").unwrap();
+    let hashed_message = Blake2b256::new(&mess_bytes);
+
+    matches!(
+        signature.verify(&verification_key, &hashed_message.as_hash_bytes()),
+        Verification::Success
+    )
+}
```
danielSanchezQ, line 46: This is the [verification proposed](https://input-output-rnd.slack.com/archives/CAUJXP739/p1638439070075400?thread_ts=1638437258.071700&cid=CAUJXP739) by @dkijania

danielSanchezQ:
Sadly this is all I got at the time, it comes from some other iterations of course. But it is as helpful as I got it. Sorry team.

PR #86, opened 2022-01-07
QA scripts

coire1:
This pull request integrates the scripts used to generate the various files and GDocs used in the QA stage in the Catalyst process.


scripts/python/qa-scripts/createVCAAggregate.py
```diff
@@ -0,0 +1,523 @@
...
+
+    def calculateOutcome(self, row):
+        bad = ''
+        good = ''
+        excellent = ''
+        tot = row[self.opt.noVCAReviewsCol]
+        if (tot >= self.opt.minimumVCA):
+            if (row[self.opt.excellentCol] > (tot/2)):
+                excellent = 'x'
+            elif (row[self.opt.notValidCol] >= (tot/2)):
```
Zeegomo, line 467: Would like to check the requirements to see why excellent requires strict majority and filtered out does not, but I cannot find it quickly now

scripts/python/qa-scripts/createVCAAggregate.py
```diff
@@ -0,0 +1,523 @@
...
+        self.allColumns = self.badColumns + self.goodColumns + self.excellentColumns
+
+        self.dChallenges = self.opt.distinctChallenges
+
+
+    def prepareBaseData(self):
+        self.gspreadWrapper.getVCAMasterData()
+        self.vcaMerged = pd.DataFrame(columns=self.gspreadWrapper.dfVca.columns, data=None)
+        self.vcaMerged['name'] = ''
+        self.dfVca = self.gspreadWrapper.dfVca.set_index('id')
```
Zeegomo, line 34: same as before, would suggest to use the appropriate opt header instead of a literal string

scripts/python/qa-scripts/createVCAAggregate.py
```diff
@@ -0,0 +1,523 @@
...
+            for aFile in currentFiles:
+                if aFile.endswith(".csv") :
+                    fpath = str(os.path.join(currentDirPath, aFile))
+                    self.vcasFiles.append(fpath)
+
+    def loadVCAsFiles(self):
+        self.prepareBaseData()
+        self.prepareVCAsFileList()
+        self.vcasData = []
+        self.vcasFileList = []
```
Zeegomo, line 54: same as before, is `vcasFileList` redundant?

scripts/python/qa-scripts/createVCAAggregate.py
```diff
@@ -0,0 +1,523 @@
...
+                    if integrity:
+                        bad = self.badFeedback(locAss)
+                        good = self.goodFeedback(locAss)
+                        excellent = self.excellentFeedback(locAss)
+                        if (self.isVCAfeedbackValid(locAss, bad, good, excellent)):
+                            if (bad or good or excellent):
+                                reviews_num_col = "No. of Reviews"
+                                if locAss[self.opt.challengeCol] in self.dChallenges:
+                                    reviews_num_col = reviews_num_col + " " + locAss[self.opt.challengeCol]
+                                if reviews_num_col in single_vca:
```
Zeegomo, line 92: How can this happen? Also this part would benefit from an explaining comment

scripts/python/qa-scripts/createVCAAggregate.py
```diff
@@ -0,0 +1,523 @@
...
+                                # Append the single review to the vcaMerged file
+                                toBeMergedAssessment = locAss.copy()
+                                toBeMergedAssessment['id'] = id
+                                toBeMergedAssessment['name'] = single_vca['name']
+                                self.vcaMerged = self.vcaMerged.append(toBeMergedAssessment)
+
+                            for col in self.allColumns:
+                                colVal = self.checkIfMarked(locAss, col)
+                                if (colVal > 0):
+                                    self.dfVca.loc[id, col] = self.dfVca.loc[id, col] + colVal
```
Zeegomo, line 106: what's this for?

scripts/python/qa-scripts/createVCAAggregate.py
```diff
@@ -0,0 +1,523 @@
...
+            self.dfVca.loc[id, 'Result ' + self.opt.excellentCol] = excellent
+
+        vcaAggregatedAssessments = pd.DataFrame(self.dfVca).copy()
+        vcaAggregatedAssessments.fillna('', inplace=True)
+        vcaAggregatedAssessments[self.opt.assessmentsIdCol] = vcaAggregatedAssessments.index
+
+        aggregatedAssessments = pd.DataFrame(self.dfVca).copy()
+        aggregatedAssessments.fillna('', inplace=True)
+        aggregatedAssessments[self.opt.assessmentsIdCol] = aggregatedAssessments.index
+        aggregatedAssessments[self.opt.excellentCol] = aggregatedAssessments['Result ' + self.opt.excellentCol]
```
Zeegomo, line 120: why don't we directly used the `self.opt.excellentCol`  from the beginning?

scripts/python/qa-scripts/createVCAAggregate.py
```diff
@@ -0,0 +1,523 @@
...
+        aggregatedAssessments[self.opt.goodCol] = aggregatedAssessments['Result ' + self.opt.goodCol]
+        aggregatedAssessments[self.opt.notValidCol] = aggregatedAssessments['Result ' + self.opt.notValidCol]
+
+        # Select valid assessments (no filtered out, no blank assessments)
+        validAssessments = pd.DataFrame(self.dfVca).copy()
+        validAssessments.fillna('', inplace=True)
+        validAssessments[self.opt.assessmentsIdCol] = validAssessments.index
+        validAssessments = validAssessments[(
+            (~(validAssessments['Result ' + self.opt.notValidCol] == 'x'))
```
Zeegomo, line 129: `not`?

scripts/python/qa-scripts/options.json.template
```diff
@@ -0,0 +1,60 @@
...
+      "6": "Auditability"
+    },
+    {
+      "5": "Impact / Alignment",
+      "3": "Feasibility",
+      "4": "Auditability"
+    }
+  ],
+  "assessmentsSheet": "Assessments",
+  "proposalKeyCol": "Idea Title",
```
Zeegomo, line 19: Could you elaborate on why is this used?

Zeegomo (commented):
Partial review,
I've noted a few things as I looked into the code but I do not expect most of the things to be fixed before merging, given the limited time and the plans for rework in the future.
I'm mostly interested in understanding the logic, code quality can be a bit ignored for now

scripts/python/qa-scripts/README.md
```diff
@@ -0,0 +1,141 @@
...
+assessments (grouping them by assessor) and we autoflag blank assessments.
+In this step the assessments made by a CA who is also a proposer in the same
+challenge will be discarded (`createProposerDocument.py`).  
+In order to run the `createProposerDocument.py` and to perform the exclusion it
+is necessary to have:
+
+    - `proposals.json` => JSON list of all proposals with `id`, `category`,
+      `title`, `url`
+    - `users.json` => JSON list of all users with `id` (the anonymized id assigned
+       to the user), `email`, `proposals` (list), `campaigns` (list), `userName`.
```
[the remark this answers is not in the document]
coire1: The proposals list is prepared manually - at the moment there isn't an automatic way to extract the connection between the anonymized assessor_id and the IdeaScale user (this is something we can automate with the ca-db).

scripts/python/qa-scripts/README.md
```diff
@@ -0,0 +1,141 @@
...
+    - In `options.json` the option `proposersMasterFile` should be set
+      with the Google Sheet key of the file generated by step `1`.
+    - In `options.json` the option `vcaMasterFile` should be set
+      with the Google Sheet key of the file generated by step `4`.
+    - In `options.json` the option `distinctChallenges` should be populated with
+      the list of the Catalyst Natives challenges. For each of this challenges
+      a specific `Valid Assessments` sheet will be generated.
+    - `proposals.json` => JSON list of all proposals with `id`, `category`,
+      `title`, `url`
+    - `vcas.json` => JSON list of all vcas with `ca_id` (the anonymized
```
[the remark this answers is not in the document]
coire1: vCA are not anonymous but they can also be CA in the same fund. The `ca_id` here is needed to associate the vCA with the anonymized CA and this relation is used later to exclude the reviews made by vCA for their own assessments.

scripts/python/qa-scripts/createProposersAggregate.py
```diff
@@ -0,0 +1,145 @@
...
+            for aFile in currentFiles:
+                if aFile.endswith(".csv") :
+                    fpath = str(os.path.join(currentDirPath, aFile))
+                    self.proposersFiles.append(fpath)
+
+    def loadProposersFiles(self):
+        self.prepareBaseData()
+        self.prepareProposersFileList()
+        self.proposersData = []
+        self.proposersFileList = []
```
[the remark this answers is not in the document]
coire1: yes, that's correct - in this file the `self.proposersFiles` and `self.proposersFileList` will always be the same. I'll update it.

scripts/python/qa-scripts/gspreadWrapper.py
```diff
@@ -0,0 +1,116 @@
...
+            # Assign triplet_id
+            df[self.opt.tripletIdCol] = df['assessor_id'] + '-' + df[self.opt.proposalIdCol].astype(str)
+            # Assign question_id
+            df[self.opt.questionIdCol] = df.groupby(self.opt.questionCol).ngroup() + 1
+            reviews = []
+            # Group assessments per triplet, to obtain the full review
+            grouped = df.groupby(self.opt.tripletIdCol)
+            for name, group in grouped:
+                for criteriaGroup in self.opt.criteriaGroups:
+                    criteriaIds = [int(i) for i in list(criteriaGroup.keys())]
```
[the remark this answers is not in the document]
coire1: In the original Ideascale export each single assessment (composed by 3 text notes and 3 ratings) is not aggregated but each row contains a text note, a rating and the question (criteria).
Combining the `assessor_id` and the `proposal_id` in the original export we can create a `tripletIdCol` and obtain the group of rows that compose a single assessment.
The total number of possible questions (criteria) in the original export is 6, divided in 2 groups (3 for standard challenges, 3 for challenge setting).
In all the files produced the 2 groups are reduced and uniformed to 1 (and then the questions/criteria to 3).
This section creates a new assessment row picking the correct values:
- from the group of rows in the original export
- for the correct group of questions (criteria)
- maintaining the correct relationship/order with the associated ratings

scripts/python/qa-scripts/createProposersAggregate.py
```diff
@@ -0,0 +1,145 @@
...
+        self.opt = Options()
+        self.utils = Utils()
+        self.gspreadWrapper = GspreadWrapper()
+        self.proposersFiles = []
+
+        self.allColumns = [self.opt.notValidCol]
+
+    def prepareBaseData(self):
+        self.gspreadWrapper.getProposersMasterData()
+        self.dfMasterProposers = self.gspreadWrapper.dfMasterProposers.set_index('id')
```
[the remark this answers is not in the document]
coire1: I agree, I'll update it.

scripts/python/qa-scripts/createProposersAggregate.py
```diff
@@ -0,0 +1,145 @@
...
+    def createDoc(self):
+        self.loadProposersFiles()
+        # Loop over master ids as reference
+        for id, row in self.dfMasterProposers.iterrows():
+            # Loop over all vca files
+            for filesIdx, proposerDf in enumerate(self.proposersData):
+                if (id in proposerDf.index):
+                    locAss = proposerDf.loc[id]
+                    integrity = self.checkIntegrity(id, row, locAss)
+                    if (integrity is False):
```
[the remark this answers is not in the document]
coire1: I agree, I'll update it.

scripts/python/qa-scripts/createProposersAggregate.py
```diff
@@ -0,0 +1,145 @@
...
+            (ass1[self.opt.q1Rating] != ass2[self.opt.q1Rating]) or
+            (ass1[self.opt.q2Rating] != ass2[self.opt.q2Rating]) or
+            (ass1[self.opt.assessorCol] != ass2[self.opt.assessorCol])
+        ):
+            return False
+        return True
+
+    def checkIfMarked(self, row, column):
+        if (row[column].strip() != ''):
+            return 1
```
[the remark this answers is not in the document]
coire1: Allowing to return an integer instead of a bool here, allows to have the final sum in the aggregated file. In case of proposers is useful to spot double submissions and for vCA is needed to calculate the consensus.

scripts/python/qa-scripts/createProposersAggregate.py
```diff
@@ -0,0 +1,145 @@
...
+    def badValid(self, row):
+        if (
+            (self.checkIfMarked(row, self.opt.notValidCol) == 1) and
+            (self.checkIfMarked(row, self.opt.notValidRationaleCol) == 0)
+        ):
+            return False
+        return True
+
+    def isProposerFeedbackValid(self, row):
+        return self.badValid(row)
```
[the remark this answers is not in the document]
coire1: It's really a bad naming :) `badValid` indicates that the calssification of a review as `bad` (or better, Filtered Out) is valid (to be valid it should include a rationale in another column).  I'll update it.

scripts/python/qa-scripts/createVCAAggregate.py
coire1: You can find the reference here: https://docs.google.com/document/d/1vHapUV1LHM4cb2375V99oM32xJ-wj_XoN-DYcvKEk0A

scripts/python/qa-scripts/createVCAAggregate.py
coire1: In this case the `vcasFileList` is not redudant because it's possible that some files from `vcasFiles` are not used (loading errors, vcas not present in the proper file).

scripts/python/qa-scripts/createVCAAggregate.py
coire1: There is the need to keep track of the reviews made by vCA for specific challenges (like Natives COTI). 
In the first part I create the base column name, that changes if the assessments is for a challenge in `dChallenges`. 
Then I update the `single_vca` dict: if it already has the key it increases the counter, otherwise it creates the key with value 1.

scripts/python/qa-scripts/createVCAAggregate.py
coire1: This is the key part of the aggregation: for each vca criteria (excellent, good, filtered out) of each assessment the counter is increased if the vCA marked it.
Using `allColumns` allows to criteria the columns without changing the logic (in previous iterations the criteria were different)

scripts/python/qa-scripts/createVCAAggregate.py
coire1: We can't because `self.opt.excellentCol` contains the count (that will be published for transparency), and `Result + ...` the outcome of the consensus based on the counters.
In some sheets published from these data both types are needed at the same time.

scripts/python/qa-scripts/createVCAAggregate.py
coire1: Yes, it's the `not` used by pandas. We use it here to filter the assessments excellent or good in the dataframe.

scripts/python/qa-scripts/options.json.template
coire1: - `allowedBlankPerAssessor` is used to check the blanks rate of CAs. In previous funds we had a max blanks ratio allowed.
- `minimumVCA` is the minimum number of vCA needed to be able to calculate consensus. In previous funds it was > 1.
- `criteriaGroups` is the map for the questions from the original IdeaScale export and the criteria columns we use.

the other values are the names of the columns used in the various sheets to extract the info needed and create the columns for each next step. 
in last part of `options.json` there are the google sheet links used in the various steps and the title of the generated docs.

scripts/python/qa-scripts/createVCAAggregate.py
Zeegomo: perfect, thanks!

scripts/python/qa-scripts/createProposersAggregate.py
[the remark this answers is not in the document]
Zeegomo: Just a nit, but you can sum boolean values just fine, and it would make reading this function a little bit easier, as you could write
```
def checkIfMarked(self, row, column):
    row[column].strip() != ''
```

scripts/python/qa-scripts/options.json.template
Zeegomo: I mean what's the purpose of the `proposalKeyCol` column? I saw that used in the scripts processing but I would expect it to be there for UX reasons only

scripts/python/qa-scripts/gspreadWrapper.py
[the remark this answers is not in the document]
Zeegomo: So just one group of the 2 is actually used for each assessment (standard/challenge) at a time?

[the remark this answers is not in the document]
coire1: correct, `filtered` is used to determine which one of the 2 groups, then I loop through the questions of the criteria to compose the final assessment.

scripts/python/qa-scripts/options.json.template
coire1: Yes, it is used only to present the name of the proposal for each assessment in the files generated.

Zeegomo: It's actually used as a group by key in https://github.com/input-output-hk/catalyst-toolbox/blob/85f7377637645b05ba9c9115af82329c354d295d/scripts/python/qa-scripts/createVCAAggregate.py#L172, that's what got me interested in this

scripts/python/qa-scripts/createVCAAggregate.py
```diff
@@ -169,13 +169,23 @@ def createDoc(self):
...
         ].mean(axis=1)
-        finalProposals = validAssessmentsRatings.groupby([self.opt.proposalIdCol, self.opt.proposalKeyCol], as_index=False)['Rating Given'].mean().round(2)
+        finalProposals = validAssessmentsRatings.groupby(
+            [self.opt.proposalIdCol, self.opt.proposalKeyCol, self.opt.challengeCol],
+            as_index=False
+        )['Rating Given'].mean().round(2)
+        fProposalsNoAssessments = validAssessmentsRatings.groupby(
+            [self.opt.proposalIdCol, self.opt.proposalKeyCol, self.opt.challengeCol],
+            as_index=False
+        ).size().reset_index()
```
Zeegomo, line 179: why is `reset_index` needed if we only read the `size` column?

scripts/python/qa-scripts/createVCAAggregate.py
```diff
@@ -169,13 +169,23 @@ def createDoc(self):
...
+        finalProposals['No. Assessments'] = fProposalsNoAssessments['size']
         for oProposal in self.proposals:
             if not (finalProposals[self.opt.proposalIdCol] == oProposal['id']).any():
                 propToAdd = {}
                 propToAdd[self.opt.proposalIdCol] = oProposal['id']
                 propToAdd[self.opt.proposalKeyCol] = oProposal['title']
+                propToAdd[self.opt.challengeCol] = ''
                 propToAdd['Rating Given'] = 0
+                propToAdd['No. Assessments'] = 0
```
Zeegomo, line 188: Are those leftover temporary impls?

PR #96, opened 2022-02-08
Exclude addresses from total stake

coire1:
This pull request enable an option in the `proposers_rewards.py` script to exclude certain addresses (e.g. committee) from the calculation of the total stake used to determine the approval threshold for proposals.

The new option refers to a local json file containing the address to be excluded in the format:
```
[
  {"address": "ca1q..."},
  {"address": "ca1q..."}
]
```

It also fixes a typo in the python dependencies and a proper sanitization for output filenames.

### DIFF
--- a/scripts/python/proposers_rewards.py
+++ b/scripts/python/proposers_rewards.py
@@ -7,6 +7,7 @@ import csv
 import itertools
 import enum
 import os
+import re
 from collections import namedtuple
 from io import StringIO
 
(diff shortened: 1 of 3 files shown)


dkijania:
Apologies i merged PR with touches `proposers_rewards.py` it was mainly enforcing types when performing threshold/stake calculation. Somehow on my machine python complains that i'm trying to compare int and str

Zeegomo (approved):

scripts/python/proposers_rewards.py
```diff
@@ -514,7 +531,12 @@ def calculate_rewards(
         )
 
         challenge_output_file_path = build_path_for_challenge(
-            output_file, challenge.title.replace(" ", "_").replace(":", "_")
+            output_file,
+            re.sub(
+                r'(?u)[^-\w.]',
```
Zeegomo, line 536: what is this regex supposed to do?

coire1: it sanitizes the file name that will be used, the filenames generated are based on the name of the challenges. We already encountered problems in win environments.

Zeegomo:
Closing this as it's already on master

PR #97, opened 2022-02-08
remove chrono as a direct dependency

Zeegomo:
Remove chrono as a direct dependency. There are still traces of it in vit-servicing-station and other dependencies, but this is a first step


Mr-Leshiy (approved):
LGTM

PR #98, opened 2022-02-09
Output reward address

Zeegomo:
This PR does the following:
* port https://github.com/input-output-hk/catalyst-toolbox/pull/75 to fund 7 branch
* ignore inactive voters from the count of total stake for the purpose of rewards
* map the rewards back to mainnet addresses (snapshot parsing not yet tested)
* introduce support for new snapshot in catalyst-toolbox

### DIFF
--- a/src/rewards/voters.rs
+++ b/src/rewards/voters.rs
@@ -175,8 +175,23 @@ mod tests {
     use chain_impl_mockchain::fee::LinearFee;
     use jormungandr_lib::crypto::account::Identifier;
     use jormungandr_lib::interfaces::{BlockchainConfiguration, Stake};
-    use quickcheck::{Arbitrary, Gen};
     use quickcheck_macros::*;
+
+    fn blockchain_configuration(initial_funds: Initial) -> Block0Configuration {
+        Block0Configuration {
+            blockchain_configuration: BlockchainConfiguration::new(
+                Discrimination::Test,
+                ConsensusVersion::Bft,
+                LinearFee::new(1, 1, 1),
+            ),
+            initial: vec![initial_funds],
+        }
+    }
+
+    #[track_caller]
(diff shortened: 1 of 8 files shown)


src/rewards/voters.rs
```diff
@@ -30,17 +42,241 @@ pub fn calculate_stake<'address>(
...
+    threshold: u64,
+) -> ActiveAddresses {
+    let discrimination = block0.blockchain_configuration.discrimination;
+    vote_count
+        .into_iter()
+        .filter_map(|(account_hex, count)| {
+            if count >= threshold {
+                Some(
+                    account_hex_to_address(account_hex, discrimination)
+                        .expect("Valid hex encoded UnspecifiedAccountIdentifier"),
```
eugene-babichenko, line 80: Is that OK to panic there? I mean, a failure with one address would mean that we got a corrupted file anyway, but I would love to have better error messaging in the future.

src/rewards/voters.rs
```diff
@@ -30,17 +42,241 @@ pub fn calculate_stake<'address>(
...
+        dbg!(&registrations);
+
+        for reg in registrations {
+            *res.entry(reg.reward_addr).or_default() +=
+                reward * Rewards::from(u64::from(reg.stake)) / total_stake;
+        }
+    }
+
+    res
+}
```
eugene-babichenko, line 133: This can be further simplified with the following algorithm:

- Sort out inactive voter keys
- Get all registrations that are above threshold for these keys
- Run the rewards distribution logic on these registrations

This would reduce the number of calculations to be done compared to the current approach and thus would make the code less error-prone.

Zeegomo: mmm, it should be what I do now

eugene-babichenko (approved):

PR #105, opened 2022-02-28
Update snapshot to support CIP-36

Zeegomo:
Add support for CIP-36 (see https://github.com/cardano-foundation/CIPs/pull/188 for more details).

~### Open problems~
~Exact format of data to be expected from Adrestia is still to be defined. The example provided lists delegations as a map, but that is in contrast with the definition of CIP-36, where the last key has special meaning.~

~This implementation refuses to parse a map since that may cause ambiguity problems and `snapshot::test_parsing` is currently failing for this reason.~

UPD: Adrestia still has to provide the specific data format but I think it's better if we go forward with this and amend the example test to use a plausible one. Use of a map would be invalid anyway as per the reason described above


src/snapshot/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+                        Delegations::New(mut vks) => {
+                            let voting_power = u64::from(voting_power);
+                            let total_weights =
+                                vks.iter().map(|(_vk, weight)| *weight as u64).sum::<u64>();
+
+                            let last = vks.pop().expect("CIP36 requires at least 1 delegation");
+                            let others_total_vp = vks
+                                .into_iter()
+                                .filter_map(|(vk, weight)| {
+                                    NonZeroU64::new((voting_power * weight as u64) / total_weights)
```
enzoc4, line 65: I don't see anything that guarantees that `total_weights` is not 0 (because all the weights can be zero I think?), but I doubt that would happen anyway

src/snapshot/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+                RawSnapshot(filtered_snapshot.clone()),
+                stake_threshold.into()
+            )
+        );
+        let mut snapshot = Snapshot::from_raw_snapshot(RawSnapshot(filtered_snapshot), 0.into());
+        snapshot.stake_threshold = stake_threshold.into();
+        assert_eq!(
+            Snapshot::from_raw_snapshot(raw, stake_threshold.into()),
+            snapshot
+        );
```
enzoc4, line 163: I don't really see the point of this assertion. The `stake_threshold` in the argument could be 0 anyway, so in theory it should be covered too?

src/rewards/voters.rs
```diff
@@ -116,20 +116,20 @@ fn rewards_to_mainnet_addresses(
...
-            .map(|reg| Rewards::from(u64::from(reg.voting_power)))
+            .map(|c| Rewards::from(c.value))
             .sum::<Rewards>();
 
-        for reg in registrations {
-            *res.entry(reg.reward_address.clone()).or_default() +=
-                reward * Rewards::from(u64::from(reg.voting_power)) / total_stake;
+        for c in contributions {
+            *res.entry(c.reward_address.clone()).or_default() +=
+                reward * Rewards::from(c.value) / total_value;
```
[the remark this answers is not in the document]
Zeegomo: the reward address might be different each time

src/snapshot/mod.rs
Zeegomo: correct, it was more of a double check that I did while writing this

src/snapshot/mod.rs
Zeegomo: well spotted

enzoc4 (approved):
I think this looks good, I'm not sure if it can be merged, because of the parsing thing, but I guess we can wait for that.

src/rewards/voters.rs
```diff
@@ -228,25 +229,68 @@ mod tests {
...
+                    .into_iter()
+                    .map(|c| c.reward_address)
+            })
+            .collect::<HashSet<_>>();
+
+        assert!(active_reward_addresses
+            .iter()
+            .all(|addr| rewards.remove(addr).unwrap() > Rewards::ZERO));
+
+        // partial test: does not check that rewards for addresses that delegated to both
```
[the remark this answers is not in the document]
Zeegomo: It's a "I don't know how to write that complete test in a reasonable way without pretty much having to rewrite the logic", so I would leave it as it is for now. If you have an idea on how to do that, please comment.

Zeegomo:
@eugene-babichenko maybe you can help with the format?

enzoc4 (approved):

src/snapshot/registration.rs
```diff
@@ -0,0 +1,320 @@
...
+                        delegations,
+                        voting_purpose: 0,
+                    }
+                })
+                .boxed()
+        }
+    }
+
+    #[test]
+    fn parse_example() {
```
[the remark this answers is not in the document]
Zeegomo: Well, json is just one of the possible formats, I wanted to add a test that could be more general

enzoc4 (approved):

PR #107, opened 2022-03-02
update jormungandr and chain-wallet-libs

dkijania:
Looks like my issue with compilation on windows is resolved. Updating jormungandr and chain-wallet-libs to newest master

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,6 +24,41 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "aead"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b613b8e1e3cf911a086f53f03bf286f52fd7a7258e4fa606f0ef220d39d8877"
+dependencies = [
+ "generic-array 0.14.5",
+]
+
+[[package]]
+name = "aes"
+version = "0.7.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
+dependencies = [
+ "cfg-if 1.0.0",
+ "cipher",
+ "cpufeatures",
+ "opaque-debug 0.3.0",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.9.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "df5f85a83a7d8b0442b6aa7b504b8212c1733da07b98aae43d4bc21b2cb3cdf6"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle 2.4.1",
+]
+
+[[package]]
 name = "ahash"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


enzoc4 (approved):

tests/tally/generator.rs
```diff
@@ -168,8 +178,8 @@ impl VoteRoundGenerator {
         };
     }
 
-    fn token_distribution(&self, ledger: &Ledger) -> TokenDistribution<()> {
-        let token_totals = TokenTotals::default()
+    fn token_totals(&self, ledger: &'a Ledger) -> TokenTotals {
```
enzoc4, line 181: is this actually necessary? you can call `token_distribution` directly on the ledger for the things this is used

tests/tally/generator.rs
```diff
@@ -181,55 +191,51 @@ impl VoteRoundGenerator {
...
+        let token_totals = self.token_totals(&tmp_ledger);
 
         self.voteplan_managers = self
             .voteplan_managers
             .clone() // cloning is cheap and make the borrowck happy
             .into_iter()
             .map(|(id, manager)| {
                 let vote_end = manager.plan().vote_end();
+                let token_distribution =
+                    TokenDistribution::new(&token_totals, tmp_ledger.accounts());
```
enzoc4, line 230:
```suggestion
                let token_distribution = tmp_ledger.token_distribution();
```

tests/tally/generator.rs
```diff
@@ -313,24 +326,20 @@ impl VoteRoundGenerator {
...
     }
 
-    pub fn statuses(&mut self) -> Vec<VotePlanStatus> {
+    pub fn statuses(&mut self, ledger: &'a Ledger) -> Vec<VotePlanStatus> {
+        let token_totals = self.token_totals(ledger);
         self.voteplan_managers
             .values()
-            .map(|manager| manager.statuses())
+            .map(|manager| {
+                manager.statuses(TokenDistribution::new(&token_totals, ledger.accounts()))
```
enzoc4, line 341:
```suggestion
                manager.statuses(ledger.token_distribution())
```

tests/tally/generator.rs
```diff
@@ -181,55 +191,51 @@ impl VoteRoundGenerator {
...
             .iter()
             .map(|member| member.public_key())
             .collect::<Vec<_>>();
 
-        let tmp_ledger = Ledger::new(block0.header().id(), block0.fragments()).unwrap();
+        let tmp_ledger = self.ledger();
         let stake_control = StakeControl::new_with(tmp_ledger.accounts(), &utxo::Ledger::new());
-        let table = chain_vote::TallyOptimizationTable::generate(stake_control.assigned().into());
+        let table = chain_vote::TallyOptimizationTable::generate(
+            NonZeroU64::new(stake_control.assigned().into()).unwrap(),
```
enzoc4, line 219: I think this is probably not correct, but I'm not sure exactly what is the right thing to do. I imagine we may want to use `batch_decrypt` here too.

This may work nonetheless if the native stake is bigger than the tokens though.

tests/tally/generator.rs
enzoc4: although, to be fair, I think this will need to removed in https://github.com/input-output-hk/chain-libs/pull/756 so it doesn't really matter that much (I mean, the argument will  not be needed anymore)

tests/tally/generator.rs
```diff
@@ -47,7 +49,7 @@ fn account_from_slice<P>(
     }
 }
 
-impl VoteRoundGenerator {
+impl<'a> VoteRoundGenerator {
```
enzoc4, line 52: I'm not sure why this annotation is needed, I think the elision rules are enough here.

tests/tally/generator.rs
dkijania: I tried to change to batch_decrypt, but keep recieving ` CannotTallyVotes { source: InvalidDecryption }'` in the test. I think this change exceeds my knowledge

PR #108, opened 2022-03-08
Move stats and qr mods from iapyx to catalyst-toolbox

dkijania:
I've developed some utils needed for tests or for business stakeholders with stats regarding active voters or total amount of ada used in voting. This PR is an attempt to move qr decoding/verifying cli command as well as snapshot/block/fragment logs statistics


src/bin/cli/kedqr/decode/hash.rs
```diff
@@ -3,17 +3,41 @@ use catalyst_toolbox::kedqr::decode;
...
 use std::{error::Error, fs::OpenOptions, io::BufReader, path::PathBuf};
 
 pub fn decode_hash(
     input: PathBuf,
     output: Option<PathBuf>,
     pin: QrPin,
 ) -> Result<(), Box<dyn Error>> {
-    // open input key and parse it
+    // generate qrcode with key and parsed pin
+    let secret = secret_from_hash(input, pin)?;
```
enzoc4, line 19: so, this seems not to be something particular to this PR since the `decode_hash` function is already there, but the name of these things is weird, why `hash`? doesn't this decrypt the QR code payload? that is by no means a hash.

dkijania: Hm, maybe hash is not very fortunate name, `decode_method` basically returns secret key based on qr payload file. Not actual image but string representation. From business perspective we differentiate payload from qr code, as user can copy payload(i call it hash) from yoroi instead of qr image from other wallets. For other cli i had to detach logic from decode_hash, since decode_hash returns file rather than secret key structure. Is renaming sufficient fix?

enzoc4: I mean, if it's terminology from yoroi or some business term maybe we should leave it. But it's certainly confusing to use _hash_ for a _ciphertext_. I mean, a hash is usually something that you don't even want to be able to decode in the first place.

dkijania: so maybe ciphertext or payload (because it's shorter) is a better name? I will rename all hash -> ciphertext or payload 

```
catalyst-toolbox kedqr decode payload 
```

or

```
catalyst-toolbox kedqr decode ciphertext 
```

dkijania:
everything should be treated as new, since there were used in test code only and review only partially

src/stats/block0/voters.rs
```diff
@@ -0,0 +1,49 @@
...
+    println!("Reading block0 from location {:?}...", block0_path);
+    let genesis = get_block(block0_path)?;
+
+    let rest_client = JormungandrRest::new(endpoint);
+    let mut stats: Stats = Default::default();
+
+    let mut total = 0;
+    for initial in genesis.initial.iter() {
+        if let Initial::Fund(initial_utxos) = initial {
+            total += initial_utxos.len();
```
Zeegomo, line 17: I don't think the initial format forbids multiple utxos for the same account

Zeegomo, line 17: I don't think the initial format forbids multiple utxos for the same account

src/stats/block0/voters.rs
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    println!("total: {}", total);
```
Zeegomo, line 21: I don't like important information from a library function to be conveyed by printing to stdout. This should be in the return value instead (see other places as well)

Zeegomo, line 21: I don't like important information from a library function to be conveyed by printing to stdout. This should be in the return value instead

src/stats/block0/voters.rs
```diff
@@ -0,0 +1,49 @@
...
+    for initial in genesis.initial.iter() {
+        if let Initial::Fund(initial_utxos) = initial {
+            for x in initial_utxos {
+                stats.total += 1;
```
Zeegomo, line 25: same consideration as above, I don't think initial utxos are guaranteed to be distinct

Zeegomo, line 25: same consideration as above, I don't think initial utxos are guaranteed to be distinct

src/stats/block0/voters.rs
```diff
@@ -0,0 +1,49 @@
...
+
+                let entry_address: chain_addr::Address = x.address.clone().into();
+                let pk = entry_address.public_key().unwrap().to_bech32_str();
+                println!("[{}/{}] Checking address state {}", stats.total, total, pk);
```
Zeegomo, line 29: some logging machinery might be better suited for a library function

Zeegomo, line 29: some logging machinery might be better suited for a library function

src/stats/block0/voters.rs
```diff
@@ -0,0 +1,49 @@
...
+                if let Ok(state) = rest_client.account_state_by_pk(&pk) {
+                    stats.obtained_voting_power += 1;
+                    if state.counters()[0] > 0 {
```
Zeegomo, line 32: this is not the correct way to test if an account has voted or not.
We have an endpoint just for that `/api/v1/votes/plan/account-votes/{account_id}` and also one for fetching this info for all accounts `/api/v1/votes/plan/accounts-votes-count`

Zeegomo, line 32: this is not the correct way to test if an account has voted or not.
We have an endpoint just for that `/api/v1/votes/plan/account-votes/{account_id}` and also one for fetching this info for all accounts `/api/v1/votes/plan/accounts-votes-count`

src/stats/block0/voters.rs
```diff
@@ -0,0 +1,49 @@
...
+                        stats.voted_at_least_once += 1;
+                    }
+                }
+            }
+        }
+    }
+
+    println!("{:?}", stats);
```
Zeegomo, line 40: same as above, a library function should not print to stdout and rather return this info

Zeegomo, line 40: same as above, a library function should not print to stdout and rather return this info

src/bin/cli/stats/block0/wallets.rs
```diff
@@ -0,0 +1,37 @@
...
+pub enum Command {
+    Count,
+    Ada,
+}
+
+impl WalletsCommand {
+    pub fn exec(&self) -> Result<(), catalyst_toolbox::stats::Error> {
+        match self.command {
+            Command::Count => {
+                calculate_wallet_distribution(&self.block0, self.threshold, self.support_lovelace)?
```
Zeegomo, line 26: I don't see how this could differ from the calculations based on snapshot

Zeegomo, line 26: I don't see how this could differ from the calculations based on snapshot

src/stats/block0/wallets.rs
```diff
@@ -0,0 +1,82 @@
...
+use jormungandr_lib::interfaces::Initial;
+use jormungandr_lib::interfaces::InitialUTxO;
+
+pub fn calculate_wallet_distribution<S: Into<String>>(
+    block0: S,
+    threshold: u64,
+    support_lovelace: bool,
+) -> Result<Stats, crate::stats::Error> {
+    let block0 = block0.into();
+    println!("Reading block0 from location {:?}...", &block0);
```
Zeegomo, line 16: no prints in lib functions please :)

Zeegomo, line 16: no prints in lib functions please :)

src/stats/block0/wallets.rs
```diff
@@ -0,0 +1,82 @@
...
+        .committees
+        .iter()
+        .cloned()
+        .map(|x| {
+            let committee_id: CommitteeId = x.into();
+            let public: chain_crypto::PublicKey<Ed25519> =
+                chain_crypto::PublicKey::from_binary(committee_id.as_ref()).unwrap();
+
+            Address(
+                "ca".to_string(),
```
Zeegomo, line 31: we should get the discrimination from block0

Zeegomo, line 31: we should get the discrimination from block0

src/stats/distribution.rs
```diff
@@ -0,0 +1,100 @@
...
+        (20_000..50_000),
+        (50_000..100_000),
+        (100_000..250_000),
+        (250_000..500_000),
+        (500_000..1_000_000),
+        (1_000_000..5_000_000),
+        (5_000_000..10_000_000),
+        (10_000_000..25_000_000),
+        (25_000_000..50_000_000),
+        (50_000_000..32_000_000_000),
```
Zeegomo, line 19: just to nitpick, this is not exhaustive and there's no guarantee in Jormungandr itself that this value will not be reached

Zeegomo, line 19: just to nitpick, this is not exhaustive and there's not guarantee in Jormungandr itself that this value will not be reached

src/stats/distribution.rs
```diff
@@ -0,0 +1,100 @@
...
+            );
+        }
+        println!(
+            "Total -> {} ",
+            self.content.values().map(|x| x.total).sum::<u64>()
+        );
+    }
+}
+
+fn format_big_number<S: Into<String>>(number: S) -> String {
```
Zeegomo, line 89: this is better done with rust format syntax https://doc.rust-lang.org/std/fmt/index.html

Zeegomo, line 89: this is better done with rust format syntax https://doc.rust-lang.org/std/fmt/index.html

src/stats/live/harvester.rs
```diff
@@ -0,0 +1,83 @@
...
+    pub fn fragment_logs(&self) -> Vec<FragmentLog> {
+        let client = reqwest::blocking::Client::builder()
+            .timeout(self.timeout)
+            .build()
+            .unwrap();
+
+        let res = client
+            .get(&format!("{}/v0/fragment/logs", self.endpoint))
+            .send()
+            .unwrap();
```
Zeegomo, line 53: this should return an error instead of unwrapping

Zeegomo, line 53: this should return an error instead of unwrapping

Zeegomo (changes requested):

src/bin/cli/kedqr/info.rs
```diff
@@ -0,0 +1,109 @@
...
+    #[structopt(short, long, parse(try_from_str))]
+    pin: QrPin,
+
+    /// Blockchain block0. Can be either url of local file path
+    #[structopt(long = "block0")]
+    pub block0: Option<String>,
+
+    /// Blockchain block0. Can be either url of local file path
+    #[structopt(short, long)]
+    pub testing: bool,
```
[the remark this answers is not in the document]
dkijania: block0 is optional, so in the case it's not provided we need to decide what discrimination to use

src/bin/cli/kedqr/verify.rs
```diff
@@ -0,0 +1,84 @@
...
+
+            let result = match self.opts {
+                QrCodeOpts::Hash => {
+                    secret_from_hash(qr_code, pin).map_err(KedQrError::KeyQrCodeHash)
+                }
+                QrCodeOpts::Img => secret_from_qr(qr_code, pin).map_err(KedQrError::KeyQrCode),
+            };
+
+            if let Err(err) = result {
+                if self.stop_at_fail {
```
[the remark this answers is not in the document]
dkijania: right

src/bin/cli/stats/archive.rs
```diff
@@ -0,0 +1,180 @@
...
+            Command::VotesByCaster => {
+                let result = archiver.number_of_votes_per_caster()?;
+                if self.calculate_distribution {
+                    self.write_to_file_or_print(ArchiveStats::calculate_distribution(&result)?)?;
+                } else {
+                    self.write_to_file_or_print(result)?;
+                }
+            }
+            Command::VotesBySlot => {
+                let result = archiver.number_of_tx_per_slot()?;
```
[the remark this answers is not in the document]
dkijania: Good point. there is a silent assumption that all transactions in archive are votes. Don't really know what should be the fix, renames `VotesBySlot` to `TxBySlot`, but then it's useless from client perspective because he doesn't know what is the subset of votes. On the other side catalyst archiver guarantees that those are votes. So maybe we should assume that all transactions in archiver are votesr and rename `number_of_tx_per_slot` to `number_of_votes_per_slot`

src/stats/block0/voters.rs
dkijania: removed command at all, since it is inefficient (due to for example DDos protection).

src/stats/block0/voters.rs
dkijania: Command removed, since there is no big business case.

dkijania: Command removed, since there is no big business case.

src/stats/block0/voters.rs
dkijania: Command removed, since there is no big business case.

dkijania: Command removed, since there is no big business case.

src/stats/block0/voters.rs
dkijania: Command removed, since there is no big business case.

dkijania: Command removed, since there is no big business case.

src/stats/block0/voters.rs
dkijania: Command removed, since there is no big business case.

dkijania: Command removed, since there is no big business case.

src/bin/cli/stats/block0/wallets.rs
dkijania: Well no always you have snapshot at hand. sometime it's just block0.

src/utils/csv.rs
```diff
@@ -22,3 +21,15 @@ pub fn dump_data_to_csv<T: Serialize>(data: &[T], file_path: &Path) -> Result<()
     }
     Ok(())
 }
+
+pub fn dump_to_csv_or_print<T: serde::Serialize + Debug>(
+    output: Option<PathBuf>,
+    result: &[T],
```
Zeegomo, line 27: better to use an iterator here so that we don't have to use stuff like `map_as_vec`

src/bin/cli/stats/archive.rs
```diff
@@ -0,0 +1,105 @@
...
+fn distribution_as_vec(
+    result: BTreeMap<std::string::String, usize>,
+) -> Result<Vec<(usize, usize)>, catalyst_toolbox::stats::Error> {
+    Ok(ArchiveStats::calculate_distribution(&result)?
+        .into_iter()
+        .map(|(x, y)| (x, y))
+        .collect())
+}
+
+fn map_as_vec(result: BTreeMap<std::string::String, usize>) -> Vec<(std::string::String, usize)> {
```
Zeegomo, line 81: this should not be needed at all

src/bin/cli/stats/live.rs
```diff
@@ -0,0 +1,50 @@
...
+            interval: self.interval,
+            logger: self.file.clone(),
+            duration: self.duration,
+        };
+
+        let harvester = Harvester::new(
+            self.endpoint.clone(),
+            std::time::Duration::from_secs(self.interval),
+        );
+        let monitor = MonitorThread::start(
```
Zeegomo, line 41: this looks overkill, why do we spawn a new thread if we don't do anything else in the meantime anyway?

src/stats/archive/calculator.rs
```diff
@@ -0,0 +1,126 @@
...
+        let mut distribution = BTreeMap::new();
+
+        for item in 0..*result
+            .values()
+            .max()
+            .ok_or(ArchiveCalculatorError::EmptyResult)?
+            + 1
+        {
+            distribution.insert(item, 0);
+        }
```
Zeegomo, line 33: it's better to use `distribution.entry(count).or_default() += 1` and remove this initialization

src/stats/archive/calculator.rs
```diff
@@ -0,0 +1,126 @@
...
+
+                for item in sorted_group.iter().skip(1) {
+                    let current: BlockDate = item.time;
+                    if are_equal_or_adjacent(&last_slot, &current, &time_era) {
+                        current_batch_size += 1;
+                    } else {
+                        max_batch_size = std::cmp::max(max_batch_size, current_batch_size);
+                        current_batch_size = 1;
+                    }
+                    last_slot = current;
```
Zeegomo, line 104: this makes a few assumptions about the underlying blockchain / selection alg, for example that it follows a FIFO order

src/stats/archive/loader.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            formatter.write_str("float in format {epoch}.{slod_id}")
+        }
+
+        fn visit_f64<E>(self, value: f64) -> Result<BlockDate, E>
+        where
+            E: serde::de::Error,
+        {
+            Ok(value.to_string().parse().unwrap())
```
Zeegomo, line 38: this should return an error instead of panicking, and btw not sure serializing blockdate as float is a good idea

Zeegomo (changes requested):

src/bin/cli/kedqr/info.rs
```diff
@@ -0,0 +1,109 @@
+use crate::cli::kedqr::decode::secret_from_hash;
+use crate::cli::kedqr::decode::secret_from_qr;
+use crate::cli::kedqr::encode::QrCodeOpts;
+use catalyst_toolbox::kedqr::QrPin;
+use chain_addr::AddressReadable;
+use chain_addr::{Discrimination, Kind};
+use chain_core::mempack::ReadBuf;
```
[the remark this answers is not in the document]
dkijania: it's outdated, currently there are in form of single import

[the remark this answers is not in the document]
Zeegomo: Ops, I don't even remember writing this comment, either it was pending in my stash and I didn't notice or GitHub messed something up. Either way ignore this, sorry

src/utils/csv.rs
```diff
@@ -12,7 +12,10 @@ pub fn load_data_from_csv<T: DeserializeOwned, const DELIMITER: u8>(
     csv_reader.deserialize().collect::<Result<Vec<T>, _>>()
 }
 
-pub fn dump_data_to_csv<T: Serialize>(data: &[T], file_path: &Path) -> Result<(), csv::Error> {
+pub fn dump_data_to_csv<'a, T: 'a + Serialize + Debug>(
```
Zeegomo, line 15: why is `Debug` needed here?

src/utils/csv.rs
```diff
@@ -22,9 +25,9 @@ pub fn dump_data_to_csv<T: Serialize>(data: &[T], file_path: &Path) -> Result<()
     Ok(())
 }
 
-pub fn dump_to_csv_or_print<T: serde::Serialize + Debug>(
+pub fn dump_to_csv_or_print<'a, T: 'a + Serialize + Debug>(
```
Zeegomo, line 28: nit: maybe we could reuse the csv format in stdout as well

Zeegomo (changes requested):

src/stats/distribution.rs
Zeegomo: This still needs to be addressed

src/stats/archive/calculator.rs
```diff
@@ -76,7 +114,7 @@ impl ArchiveStats {
     pub fn max_batch_size_per_caster(
         &self,
         slots_in_epoch: u32,
-    ) -> Result<BTreeMap<String, usize>, ArchiveCalculatorError> {
+    ) -> Result<Vec<CasterBatchSize>, ArchiveCalculatorError> {
```
Zeegomo, line 117: same here, may `caster_id` appear twice in the `Vec` with different meanings? Otherwise a map is more idiomatic

Zeegomo (changes requested):

src/utils/csv.rs
dkijania: not neede. removing

PR #110, opened 2022-03-10
Define voting HIR format

Zeegomo:
Define High Level Intermediate Representation (HIR) for voting entities in the Catalyst ecosystem.
This is intended as a high level description of the setup, which is not enough on its own to spin a blockchain, but it's slimmer, easier to understand and free from implementation constraints.
You can roughly read this as "voting_key will participate in this voting round with role voting_group and will have voting_power influence"

This builds on the concept of voting groups as described in https://github.com/input-output-hk/vit-servicing-station/pull/198

This is an intermediate step in the proposed process of spinning up a blockchain from the raw mainnet snapshot.

Such process is planned to look as follow:
* collect valid catalyst registrations (db-sync + Adrestia's voting tools)
* collect valid representative registrations (TBD)
* aggregate all registrations and produce a single set of participants in the voting event ( snapshot tool in this repo)
* setup blockchain parameters (block0, ...) and spin up nodes (vitup)

Not sure if catalyst toolbox is actually the correct place to host this.


voting-hir/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+/// You can roughly read this as
+/// "voting_key will participate in this voting round with role voting_group and will have voting_power influence"
+pub struct VotingHIR {
+    pub voting_key: Identifier,
+    /// Voting group this key belongs to.
+    /// If this key belong to multiple voting groups, multiple records for the same
+    /// key will be used.
+    pub voting_group: VotingGroup,
+    /// Voting power as processed by the snapshot
+    pub voting_power: Value,
```
Mr-Leshiy, line 21: Am I right that `voting_power` is referenced to the `voting_key` ?

voting-hir/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+/// entities in the Catalyst ecosystem.
+///
+/// This is intended as a high level description of the setup, which is not
+/// enough on its own to spin a blockchain, but it's slimmer, easier to understand
+/// and free from implementation constraints.
+///
+/// You can roughly read this as
+/// "voting_key will participate in this voting round with role voting_group and will have voting_power influence"
+pub struct VotingHIR {
+    pub voting_key: Identifier,
```
Mr-Leshiy, line 15: Is `voting_key` a user Identifier ? If so maybe it is better to rename it to `users_voting_key` to reflect that it belongs to user.
Otherwise I think we should provide a description does it mean.

voting-hir/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+    /// Voting group this key belongs to.
+    /// If this key belong to multiple voting groups, multiple records for the same
```
Mr-Leshiy, line 17: But why does not to store it inside an array if we have multiple records ?

voting-hir/src/lib.rs
Zeegomo: Yes

voting-hir/src/lib.rs
Zeegomo: I don't get what you are suggesting

voting-hir/src/lib.rs
Zeegomo: I mean, who should this voting key belong to otherwise? If we rename the struct to `VoterHIR` would this be clearer?

voting-hir/src/lib.rs
Mr-Leshiy: I am talking about to replace `pub voting_group: VotingGroup,` to the `pub voting_group: Vec<VotingGroup>,`, so to not create a one record for the `voting_key`

voting-hir/src/lib.rs
Mr-Leshiy: Yes, it looks better to me

voting-hir/src/lib.rs
Zeegomo: but those voting groups might have different voting power, so you should also have a `Vec` of voting power, but then the relation between the two is not so tight anyway. Also, in the beginning we're probably only going to have one group per voter anyway, and this seems like the simplest solution to ma that is both not an overkill for now and flexible enough in the future.

voting-hir/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Define High Level Intermediate Representation (HIR) for voting
+/// entities in the Catalyst ecosystem.
+///
+/// This is intended as a high level description of the setup, which is not
+/// enough on its own to spin a blockchain, but it's slimmer, easier to understand
+/// and free from implementation constraints.
+///
+/// You can roughly read this as
+/// "voting_key will participate in this voting round with role voting_group and will have voting_power influence"
+pub struct VoterHIR {
```
enzoc4, line 14: are we going to implement serde traits for this?

Zeegomo: Yes, even though we may want to serialize the voting key differently from the default bech32 encoding (e.g. hex). Maybe I should add some helper methods for that here

Zeegomo:
superseded by https://github.com/input-output-hk/catalyst-toolbox/pull/129

PR #112, opened 2022-03-17
migrate from py03 to standard cmd for proposers_rewards

dkijania:


Cargo.toml
```diff
@@ -70,15 +70,13 @@ chain-vote = { git = "https://github.com/input-output-hk/chain-libs.git", branch
...
 features = ["auto-initialize"]
 optional = true
 
-
-[build-dependencies]
-versionisator = "1.0.3"
-
-
 [features]
 test-api = ["pyo3"]
```
Zeegomo, line 82: I think we can remove this

src/testing.rs
```diff
@@ -13,12 +17,19 @@ pub struct ProposerRewardsExecutor {
...
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+}
+
+impl Default for ProposerRewardsCommand {
     fn default() -> Self {
         Self {
+            python_exec: "python3".to_string(),
```
Zeegomo, line 32: This requires python3 to be in path, not sure of how common that is, but I would at least put a disclaimer somewhere that it's needed for testing

Zeegomo (approved):

PR #116, opened 2022-03-17
switch stake calculations for rewards from funds to tokens

dkijania:


src/rewards/voters.rs
```diff
@@ -27,8 +27,11 @@ fn calculate_active_stake<'address>(
 
     for fund in &block0.initial {
         match fund {
-            Initial::Fund(fund) => {
-                for utxo in fund {
+            Initial::Fund(_) => {}
+            Initial::Cert(_) => {}
+            Initial::LegacyFund(_) => {}
+            Initial::Token(token) => {
+                for utxo in &token.to {
```
Zeegomo, line 34: utxo is a bit misleading here

Zeegomo (approved):
Should this go in for fund8?

Cargo.toml
```diff
@@ -27,11 +27,11 @@ chain-time = { git = "https://github.com/input-output-hk/chain-libs.git", branch
 chain-impl-mockchain = { git = "https://github.com/input-output-hk/chain-libs.git", branch = "master" }
 time = { version = "0.3", features = ["formatting", "parsing", "macros"] }
 itertools = "0.10"
-jcli = { git = "https://github.com/input-output-hk/jormungandr.git", branch = "master" }
-jormungandr-lib = { git = "https://github.com/input-output-hk/jormungandr.git", branch = "master" }
-jormungandr-integration-tests = { git = "https://github.com/input-output-hk/jormungandr.git", branch = "master" }
-jormungandr-automation = { git = "https://github.com/input-output-hk/jormungandr.git",  branch = "master" }
-thor = { git = "https://github.com/input-output-hk/jormungandr.git",  branch = "master" }
+jcli = { git = "https://github.com/input-output-hk/jormungandr.git", branch = "update_jortestkit" }
```
Zeegomo, line 30: is this needed?

dkijania: it was needed, because in vit-testing we got a conflict (jormungandr has to be adjust before updating jortestkit), since now this branch is merged, we can discard this change

Zeegomo: This might be a problem for the fund8 release, since we will have to rely on the jormungandr fund8 branch which does not have this change (we'll likely have to backport the jortestkit update), but I think it would be better to use release branches instead.

src/snapshot.rs
```diff
@@ -22,6 +22,12 @@ pub struct CatalystRegistration {
 #[derive(Deserialize, Clone, Debug)]
 pub struct RawSnapshot(Vec<CatalystRegistration>);
 
+impl From<Vec<CatalystRegistration>> for RawSnapshot {
+    fn from(from: Vec<CatalystRegistration>) -> Self {
+        Self(from)
+    }
+}
+
```
Zeegomo, line 30: Is this needed somewhere?

dkijania: yes in vit testing

PR #119, opened 2022-03-23
add command to output block0 initials

Zeegomo:


enzoc4 (approved):
LGTM

src/bin/cli/snapshot/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+    OutputFile(#[from] OutputFileError),
+    #[error(transparent)]
+    OutputFormat(#[from] OutputFormatError),
+}
+
+impl SnapshotCmd {
+    pub fn exec(self) -> Result<(), Error> {
+        let raw_snapshot: RawSnapshot = serde_json::from_reader(File::open(&self.snapshot)?)?;
+        let initials = Snapshot::from_raw_snapshot(raw_snapshot, self.threshold)
+            .to_block0_initials(self.discrimination);
```
dkijania, line 52: What vitup expects is that root of json is initials, like below:

```
{
    "initial": [
        {
            "fund": [
                {
```

But this is not a big problem. However there is another discrepancy as funds collections are split by 100 items. Question if not splitting is ok? i remember that there was some upper limit of items in fund

Zeegomo: I guess that vitup expects a `Block0Configuration`, while here the only thing provided is the list of initials, so I would keep it that way to make it explicit that this tools only provides that.

For the second issue, I think that has to do with limits in fragment size, i.e. it's not logically wrong to have all fragments in a single `Initial::Fund`, but only 255 will fit in a fragment on chain. Tbh, we do implement `impl TryFrom<Initial> for Fragment`, so there is some expectation that you should be able to convert between the two.

If it's not too much work on vitup, I would suggest we keep this tool as blockchain agnostic as possible, and change this to return just a `Vec<InitialUTxO>` (i.e. just a collection of inital funds), leaving splitting to deployment tools.

src/bin/cli/snapshot/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+        let mut out_writer = self.output.open()?;
+        let content = self
```
Mr-Leshiy, line 54: maybe better to change places
```
let mut out_writer = self.output.open()?;
```
and 
```
let content = self.output_format
             .format_json(serde_json::to_value(initials)?)?;
``` 
So if some errors appears `format_json(serde_json::to_value(initials)?)` we will not even open output file

Zeegomo: We would close it right after an error anyway and I don't expect `serde_json::to_value(initials)` to ever fail, so I don't think it makes much of a difference

src/bin/cli/snapshot/mod.rs
dkijania: Agree!

dkijania (approved):

PR #132, opened 2022-05-11
Filter dissenting vCAs

cameron1024:
Provides a CLI option to filter vCAs who disagree with the most common opinion: `catalyst-toolbox ideascale filter`

Open questions:
 - is `catalyst-toolbox ideascale filter` the right command?
 - should it be merged into another command (maybe it happens automatically?)
 - what happens in the case where the most common opinion is "filter", do we:
   - filter out all the `good` and `excellent`votes and leave the `filtered_out` votes OR
   - remove all the votes for that proposal/CA pair
  - what happens in the case of ties/non-absolute majorities. The current behaviour is a simple mode. What happens in the case of a 30:30:40 split?


minikin:
@cameron1024 Can you fix conflicts, please?

minikin:
@Mr-Leshiy @saibatizoku please, review.

saibatizoku (approved):

PR #133, opened 2022-05-12
Add active voter stats from vote count

dkijania:
This PR is implementation for requests from Research team for additional stats:

Request 1:

```
Could we expand following levels please - I was asked if we can be more granular on small.

450-500
[500-1000]
[1000-200]
[2000-5000]
[5000-10000]
```

Request 2: 

```
is there a way to find out how many individual votes were cast per wallet basis?
we are trying to establish how many votes do different sized wallets tend to cast
do bigger whales care more about proposals
do smaller wallets cast more votes - or if there is no correlation
```

Request 3:
```
Are we able to parse data based on votes cast range?
Meaning number of wallets with a-b votes cast something like this
1-5
5-10
10-20
20-50
50-100
100-200
200-400
400-800
800+

So above is total votes cast - and how many wallets did that
```

### DIFF
--- a/catalyst-toolbox/src/stats/voters.rs
+++ b/catalyst-toolbox/src/stats/voters.rs
@@ -1,22 +1,20 @@
+use crate::rewards::voters::{account_hex_to_address, VoteCount};
 use crate::stats::distribution::Stats;
 use chain_addr::{Discrimination, Kind};
 use chain_crypto::Ed25519;
 use chain_impl_mockchain::vote::CommitteeId;
 use jormungandr_automation::testing::block0::get_block;
 use jormungandr_lib::interfaces::Address;
+use jormungandr_lib::interfaces::Block0Configuration;
 use jormungandr_lib::interfaces::Initial;
 use jormungandr_lib::interfaces::InitialUTxO;
+use std::path::Path;
 
-pub fn calculate_wallet_distribution<S: Into<String>>(
-    block0: S,
-    threshold: u64,
-    support_lovelace: bool,
-) -> Result<Stats, crate::stats::Error> {
-    let block0 = block0.into();
-    let genesis = get_block(block0)?;
+fn blacklist_addresses(genesis: &Block0Configuration) -> Vec<Address> {
+    let discrimination = genesis.blockchain_configuration.discrimination;
 
     #[allow(clippy::needless_collect)]
-    let blacklist: Vec<Address> = genesis
+    genesis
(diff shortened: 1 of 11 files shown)


catalyst-toolbox/src/stats/voters.rs
```diff
@@ -36,21 +33,86 @@ pub fn calculate_wallet_distribution<S: Into<String>>(
...
+                        .unwrap()
+                            == utxo.address
+                    }) {
+                        return Some((utxo.clone(), *votes_count as u32));
+                    }
+                }
+            }
+            None
+        })
+        .collect()
```
Zeegomo, line 62: Reversing this iteration would be much faster. Start with initials and see for each one if it's in `VoteCount`.

catalyst-toolbox/src/stats/voters.rs
```diff
@@ -1,22 +1,20 @@
...
-    block0: S,
-    threshold: u64,
-    support_lovelace: bool,
-) -> Result<Stats, crate::stats::Error> {
-    let block0 = block0.into();
-    let genesis = get_block(block0)?;
+fn blacklist_addresses(genesis: &Block0Configuration) -> Vec<Address> {
+    let discrimination = genesis.blockchain_configuration.discrimination;
 
     #[allow(clippy::needless_collect)]
```
Zeegomo, line 16: we can remove this now I think

catalyst-toolbox/src/stats/voters.rs
```diff
@@ -36,21 +32,87 @@ pub fn calculate_wallet_distribution<S: Into<String>>(
...
+                for utxo in funds {
+                    if let Some((_, votes_count)) = votes_count.iter().find(|(address, _)| {
+                        account_hex_to_address(
+                            address.to_string(),
+                            genesis.blockchain_configuration.discrimination,
+                        )
+                        .unwrap()
+                            == utxo.address
+                    }) {
+                        return Some((utxo.clone(), *votes_count as u32));
```
Zeegomo, line 56: the iteration through `votes_count` is still a bit inefficient, I think we can go from the address to the hex key and access `votes_count` as a map directly  e.g.
```
for utxo in funds {
    if let Some(vote_count) = votes_count.get(addr_to_hex(utxo.address)) {
        return Some(utxo.clone(), vote_count)
    }
}
```

Mr-Leshiy (approved):
LGTM

PR #135, opened 2022-05-18
Add voting power cap during snapshot

Zeegomo:
Limit overall influence of actors in an election by capping their
maximum voting power at arbitrary levels.

Since a lot of configurations are going in the snapshot process,
I'm thinking of having it output two different artifacts:
the list of VoterHIRs, as already proposed, and another one which
is a light wrapper around VoterHIR which contains contributions
from snapshot, to be used for the reward process.

I'm proposing to keep the separate because the first one is
more general, while the second one is tied to our current
deployment and more likely to change.

### Unimplemented

The rewards cmd is going to change and I did not put too much effort there to update it. My idea was to not re-run the snapshot cmd again each time but instead use the format I proposed above, which would be automatically created in the snapshot pipeline.

### DIFF
--- a/catalyst-toolbox/src/bin/cli/rewards/voters.rs
+++ b/catalyst-toolbox/src/bin/cli/rewards/voters.rs
@@ -5,9 +5,9 @@ use catalyst_toolbox::snapshot::{
 use catalyst_toolbox::utils::assert_are_close;
 
 use color_eyre::Report;
+use fraction::Fraction;
 use jcli_lib::jcli_lib::block::Common;
 use jormungandr_lib::{crypto::account::Identifier, interfaces::Block0Configuration};
-
 use structopt::StructOpt;
 
 use std::collections::BTreeMap;
--- a/catalyst-toolbox/src/bin/cli/rewards/voters.rs
+++ b/catalyst-toolbox/src/bin/cli/rewards/voters.rs
@@ -1,10 +1,12 @@
 use catalyst_toolbox::rewards::voters::{calc_voter_rewards, Rewards, VoteCount};
-use catalyst_toolbox::snapshot::{registration::MainnetRewardAddress, Snapshot};
+use catalyst_toolbox::snapshot::{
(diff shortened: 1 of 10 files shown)


catalyst-toolbox/src/snapshot/influence_cap.rs
```diff
@@ -0,0 +1,200 @@
...
+    (a + b - 1) / b
+}
+
+/// Let's denote with X the current voting power of user U and with TOT the total voting power
+/// and suppose X / TOT >= threshold.
+/// We want to remove a quantity Y from X such that (X - Y)/ (TOT - Y) <= treshold.
+/// The solution to this equation is therefore Y <= (X - T*TOT) / (1 - T).
+/// Since we work with integer numbers, we will use Y' = ceil(Y).
+///
+/// Panics if threshold is >= 1 or x/tot <= threshold
```
Mr-Leshiy, line 18: I think `threshold` value could be equal to `1`.
Need to change `threshold is > 1`

Zeegomo: yep, the comment is outdated

Mr-Leshiy (approved):

catalyst-toolbox/src/bin/cli/rewards/voters.rs
```diff
@@ -92,3 +101,10 @@ impl VotersRewards {
         Ok(())
     }
 }
+
+struct DummyAssigner;
+impl VotingGroupAssigner for DummyAssigner {
+    fn assign(&self, _vk: &Identifier) -> String {
+        unimplemented!()
```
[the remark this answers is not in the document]
minikin: @Zeegomo @Mr-Leshiy btw for such cases we can adapt something like this 
```txt
// TODO(minikin): This ignore shouldn't be necessary, see https://github.com/rust-lang/rust/issues/97932
```

enzoc4 (approved):

PR #148, opened 2022-07-07
[Fund9] Revert unneeded changes

Zeegomo:
Some of the changes originally planned for fund9 has been delayed to fund10, which means that some of the already in progress work has to be removed from the fund9 branch, which is exactly what this PR does.

Other ergonomics changes (e.g. better error reporting) or bug fixes remain untouched


cameron1024 (approved):
LGTM

PR #152, opened 2022-07-21
make VoterHIR public

Mr-Leshiy:
This update is needed for the vit-servising-station modification.
https://github.com/input-output-hk/vit-servicing-station/pull/254


Zeegomo (changes requested):
can't you simply use the `voting-hir` crate from wherever you need it?

Zeegomo (approved):

PR #156, opened 2022-08-01
Implement --minimum_consensus parameter for vca reward command

2072:
Value in range [0.5, 1]
The minimum consensus for a vCA ranking to be excluded from eligible rankings when in disagreement from simple majority.
Simple majority is 50%.
Qualified majority is 70%. Using 70% avoids punishing vCAs where the consensus is not clear.
70% is because when #vca == 3 consensus is only 66% and thus in this case, where there is just 1 vote in disagreement, all 3 vCAs get rewarded.


2072:
Here is a spreadsheet showing the difference of output: https://docs.google.com/spreadsheets/d/1Sam0fcEIRdx7flujAiKzS0zcM60UfdQupJBy8altbzw/edit#gid=0

Note that I used the f8-vca-merged.csv that was removed a few commit ago.

2072:
I've updated the spreadsheet and also added the output of the original tool without this patch into a new tab to show that it is identical when using `--minimum_consensus 0.5`.
Note that this is the first time I deal with Rust so do not hesitate to tell me if there are better ways,,,

Zeegomo:
Hey @2072, thank you for your PR!
We do already filter out from rewards advisors in disagreement with the majority in https://github.com/input-output-hk/catalyst-toolbox/blob/49e4a02a3f3e009bd078745a782c7ac10aa03861/catalyst-toolbox/src/rewards/veterans.rs#L53 (exact threshold is tunable from cli).
 Or is what you are proposing different from that?

Also, I'm lacking a bit of context, is there a broader consensus behind this change or is it just your (valid) suggestion?

catalyst-toolbox/src/rewards/veterans.rs
```diff
@@ -77,13 +77,35 @@ fn calc_final_eligible_rankings(
         .collect()
 }
 
+fn calc_final_ranking_consensus_per_review(rankings: &[impl Borrow<VeteranRankingRow>]) -> Decimal {
```
Zeegomo, line 80: let's not duplicate code, this can be done together with `calc_final_ranking_per_review` as a sort of score confidence

catalyst-toolbox/src/rewards/veterans.rs
```diff
@@ -77,13 +77,35 @@ fn calc_final_eligible_rankings(
...
+        (Some(filtered_out), _, _) if Decimal::from(*filtered_out) >= rankings_majority => {
+            Decimal::from(*filtered_out) / Decimal::from(rankings.len())
+        }
+        (_, Some(excellent), _) if Decimal::from(*excellent) > rankings_majority => {
+            Decimal::from(*excellent) / Decimal::from(rankings.len())
+        }
+        (_, Some(excellent), Some(good)) => {
+            (Decimal::from(*excellent) + Decimal::from(*good)) / Decimal::from(rankings.len())
+        }
+        (_, _, Some(good)) => {
```
Zeegomo, line 94: nit: this can be grouped with the arm above by doing
```
(_, maybe_excellent, Some(good)) => Decimal::from(maybe_excellent.unwrap_or_default()) + ...
```

catalyst-toolbox/src/rewards/veterans.rs
```diff
@@ -77,13 +77,35 @@ fn calc_final_eligible_rankings(
...
+           Decimal::from(*good) / Decimal::from(rankings.len())
+        }
+        _ => Decimal::ONE,
```
Zeegomo, line 97: not sure this arm can ever be reached and I'm a bit confused by the fact we're using 1 here as a value

catalyst-toolbox/src/bin/cli/rewards/veterans.rs
```diff
@@ -62,6 +62,14 @@ pub struct VeteransRewards {
     /// if the first cutoff is selected then the first modifier is used.
     #[structopt(long, required = true)]
     reputation_agreement_rate_modifiers: Vec<Decimal>,
+
+    /// Value in range [0.5, 1]
+    /// The minimum consensus for a vCA ranking to be excluded from eligible rankings when in disagreement from simple majority.
+    /// Simple majority is 50%.
+    /// Qualified majority is 70%. Using 70% avoids punishing vCAs where the consensus is not clear.
+    /// 70% is because when #vca == 3 consensus is only 66% and thus in this case, where there is just 1 vote in disagreement, all 3 vCAs get rewarded.
+    #[structopt(long = "minimum_consensus")]
```
Zeegomo, line 71: nit: we use kebap-case convetion for our names, it's enough to just tag this  `#[structopt(long)]`

catalyst-toolbox/src/bin/cli/rewards/veterans.rs
```diff
@@ -113,6 +126,7 @@ impl VeteransRewards {
                 .into_iter()
                 .zip(reputation_agreement_rate_modifiers.into_iter())
                 .collect(),
+            Decimal::from(minimum_consensus),
```
Zeegomo, line 129: unneeded conversion

Zeegomo (changes requested):
Good work! I left some comments

catalyst-toolbox/src/rewards/veterans.rs
2072: I had to do a map to transform our `maybe_excellent`  (Option[&usize]) into an  Option[Decimal] before I could use the unwrap_or_default() on it:

```Rust
(_, maybe_excellent, Some(good)) => {
            (maybe_excellent.map(|excellent| Decimal::from(*excellent)).unwrap_or_default() + Decimal::from(*good)) / Decimal::from(rankings.len())
        }
```

As I said I'm not familiar with Rust at all so there might be a better way... (I'm very used to Scala)

catalyst-toolbox/src/bin/cli/rewards/veterans.rs
2072: fixed

catalyst-toolbox/src/rewards/veterans.rs
2072: ok thanks! fixed

catalyst-toolbox/src/rewards/veterans.rs
2072: OK, I'm going to try, good Rust exercise

catalyst-toolbox/src/rewards/veterans.rs
2072: fixed by 74a5544, tell me if there is a better way

catalyst-toolbox/src/rewards/veterans.rs
2072: ok done in 6d47932

2072:
OK, all your requested changes are done, this allowed me to caught a small issue on the consensus of Excellent assessment where the number of Good where not taken into account (very low impact on rewards du to the rareness of Excellent and the rareness of positive disagreement on those).

catalyst-toolbox/src/rewards/veterans.rs
```diff
@@ -25,18 +30,34 @@ pub type EligibilityThresholds = std::ops::RangeInclusive<usize>;
...
+                consensus: Decimal::from(*filtered_out) / Decimal::from(rankings.len()),
+            }
         }
-        (_, Some(excellent)) if Decimal::from(*excellent) > rankings_majority => {
-            ReviewRanking::Excellent
+        (Some(excellent), maybe_good, _) if Decimal::from(*excellent) > rankings_majority => {
+            FinalRankingWithConsensus {
+                review_ranking: Excellent,
+                consensus: (maybe_good.map(|good| Decimal::from(*good)).unwrap_or_default()
+                    + Decimal::from(*excellent)) / Decimal::from(rankings.len()),
```
Zeegomo, line 50: In a way this interpretation is a bit different from what I would expect. We are promoting `good` votes to excellent as far as consensus goes but those were for a strictly lower level (i.e. if I voted for it to be good probably I think it's not worth of an excellent vote).
The opposite is fine ofc, if I voted for it to be excellent I would obviously be fine with it being good.

It's fine for me as long as everybody agrees on this, and if we go this way we might as well collapse these 2 match arms together

catalyst-toolbox/src/rewards/veterans.rs
Zeegomo: ah I see, `Option` of a reference type has no default, but this should work
```
(_, maybe_excellent, Some(good)) => Decimal::from(maybe_excellent.copied().unwrap_or_default()) + ...

```
`copied()` turns `Option<&usize>` into `Option<usize>` by copying the value

catalyst-toolbox/src/rewards/veterans.rs
```diff
@@ -217,23 +215,39 @@ mod tests {
...
+            calc_final_ranking_with_consensus_per_review(&gen_dummy_rankings("".into(), 4, 1, 5, RandomIterator)),
+            FinalRankingWithConsensus {
+                review_ranking: FilteredOut,
+                consensus,
+            } if consensus == (dec!(5) / dec!(10))
         ));
 
+        let excellent_final =calc_final_ranking_with_consensus_per_review(&gen_dummy_rankings("".into(), 3, 1, 1, RandomIterator));
+
+            dbg!(excellent_final.consensus);
```
Zeegomo, line 244: leftover?

Zeegomo (changes requested):

catalyst-toolbox/src/rewards/veterans.rs
2072: I agree with you on the principle but I did it for consistency given that we do not discriminate between `Good` and `Excellent` anywhere else in the code, not doing this would keep `FO` ratings that should not be kept when the final rating is `Excellent`:

Since the current intent behind the `consensus` value is to prevent removing disagreeing ratings in a tie, only counting the `Excellents` to compute the consensus of an `Excellent` rating would decrease the consensus of the overall positivity of the `Excellent` rating and prevent excluding the FO ratings on those `Excellents`.

For example, in the case of a (4,1,1) we would have a consensus of 4/6 (<.70) which means that he vCA who chose `FO` would still be rewarded for their `FO` rating (if minimum consensus is set to 70%) whereas if we include the `Goods` in the consensus we get a consensus of 5/6 (0.83) and then the `FO` rating is no longer legitimate anymore (the `final_ranking_with_consensus.consensus < minimum_consensus` condition is `false` and the assessment is not kept).

This is what I've tried to explain in the comment above the function but if we wanted to be thorough the 'consensus' value should be renamed to something like `conseus_on_negativity_or_positivity` (maybe I should do it?). It's true that for now we just test for `.is_positive()` but this might change in the future so I think it's important to keep the two match arms.

Zeegomo: I understand your point, it just feels like there are a lot of very particular rules which make understanding the system as a whole more difficult than it should IMO (this is not on you ofc)

catalyst-toolbox/src/rewards/veterans.rs
2072: fixed by c5216d1

catalyst-toolbox/src/rewards/veterans.rs
2072: I've renamed `consensus` to `negative_or_positive_consensus` and documented it in b4e3b49

catalyst-toolbox/src/rewards/veterans.rs
2072: thanks, it's fixed in 3917c88

catalyst-toolbox/src/rewards/veterans.rs
```diff
@@ -16,6 +16,15 @@ pub struct VeteranAdvisorIncentive {
     pub reputation: u64,
 }
 
+pub struct FinalRankingWithConsensus {
```
Zeegomo, line 19: this does not need to be `pub`, same for fields

catalyst-toolbox/src/rewards/veterans.rs
```diff
@@ -16,6 +16,15 @@ pub struct VeteranAdvisorIncentive {
...
+    pub review_ranking: ReviewRanking,
+
+    /// This is to be used in conjunction with `ReviewRanking::is_positive()` to assess the
+    /// consensus of the boolean reply. It is either `#FO / #Rankings` or `(#Excellent + #Good) / #Rankings`.
+    /// For now we do not discriminate between Good and Excellent but this might change in the future.
+    pub negative_or_positive_consensus: Decimal,
```
Zeegomo, line 25: Actually I find this name quite confusing, `consensus_confidence` or something in the likes (or even `consensus` like before) is better

Zeegomo (changes requested):
Last few things

catalyst-toolbox/src/rewards/veterans.rs
2072: ok, it was to clearly link it to the `is_positive()` because with just `consensus` there is this confusion that it's the actual consensus of the final rating while it's not (except when it's `FO`) but the documentation I added might be enough to avoid this confusion.

Zeegomo: > there is this confusion that it's the actual consensus of the final rating while it's not

It is in a way, with the additional difference explained in the comment. IMO naming it confidence would remove this problem since a lower score is just a general indication that not everybody agreed on the ranking without delving too much into the details

catalyst-toolbox/src/rewards/veterans.rs
2072: fixed

catalyst-toolbox/src/rewards/veterans.rs
2072: Ok I agree. fixed by a68edef

btw, I'm not used to this GitHub review procedure: should I resolve myself each conversation like I did until now or just reply and let you resolve them (so that you can agree that it is indeed resolved)?
Do I have to re-request the review after my commit or is it not necessary?

2072: and [63ba8c9](https://github.com/input-output-hk/catalyst-toolbox/pull/156/commits/63ba8c9a113c960e495e4f4c43fd199e0fd3ae5d) replace the term `consensus` with `confidence` everywhere else.

Zeegomo: > should I resolve myself each conversation like I did until now or just reply and let you resolve them (so that you can agree that it is indeed resolved)?

It's fine as you did in these threads, I'm going to take a look anyway and add comments if I think there's more to add


> Do I have to re-request the review after my commit or is it not necessary?

Not necessary but if I forget to get back to you it's a good reminder

Zeegomo (approved):
Good work

2072:
Thanks!

> Good work

Thanks!

PR #157, opened 2022-08-04
Use integer values for rewards input/output

Zeegomo:
Internal calculations are still done using decimal values for
better precision but input, and mainly output, are now integer
values to better support lovelaces calculation (or actually any
other token).


Mr-Leshiy (approved):
LGTM

PR #159, opened 2022-08-09
Create new crate snapshot-lib

Mr-Leshiy:
This PR is intended to remove a cyclic dependency with the `vit-servicing-station`.


saibatizoku (approved):
Thanks for refactoring and helping to reduce cyclic dependencies!

